Issue #100, opened 2023-09-29
Create action button / ARPA button component


maxatdetroit:
**Describe the feature you'd like**

Build an action button / ARPA button component in the UXDS that is configurable to meet current current uses of action buttons / ARPA buttons.

Additional features to include:
* ARPA buttons should include an icon or other indicator that signifies they are buttons on both desktop and mobile

![Screenshot 2023-09-29 at 4 09 52 PM](https://github.com/CityOfDetroit/COD-Design-System/assets/143553259/9fd3f0f5-4861-489a-898e-7578342d6d32)
![Screenshot 2023-09-29 at 4 10 17 PM](https://github.com/CityOfDetroit/COD-Design-System/assets/143553259/5cfc66db-e575-4cc2-a3c4-c71cd1221348)
![arpa-button-1](https://github.com/CityOfDetroit/COD-Design-System/assets/143553259/4d581c5c-0dad-4569-8672-c32b00ccca1e)
![arpa-button-2](https://github.com/CityOfDetroit/COD-Design-System/assets/143553259/66c793da-6ffb-4877-a800-2862c675f4ae)
![action-card-1](https://github.com/CityOfDetroit/COD-Design-System/assets/143553259/5978b2ef-f065-43c9-80a8-336a517bbc67)

**Describe alternatives you've considered**

Alternatives:
* Implementing action button & ARPA button as separate components. This may still be desirable. I'll have to look into the code and make a judgement call.

**Additional context**

See examples of current ARPA buttons here:
* https://detroitmi.gov/departments/office-chief-financial-officer/how-detroits-arpa-funds-are-being-spent

maxatdetroit:
@bankovic if you have more examples of ARPA card flavors, would you post pictures here?

maxatdetroit:
More examples after discussing with @bankovic: 
* https://detroitmi.gov/departments/office-chief-financial-officer/how-detroits-arpa-funds-are-being-spent
* https://detroitmi.gov/departments/construction-and-demolition-department
* https://detroitmi.gov/departments/detroit-department-transportation
* https://detroitmi.gov/departments/civil-rights-inclusion-opportunity-department

Specific elements/features to include:
* Primary header
* Secondary header
* Primary background
* Secondary background
* Icon
* (Optional) top image
* (Optional) body text

PR #103, opened 2023-10-05
Support action buttons in UXDS

maxatdetroit:
Part of #100 

## This PR

1. Creates a icon container for using icons within cards
2. Creates a higher-order element `<clickable/>` that causes child elements to behave as a clickable element
3. Adds funtionality to the card element to become a clickable component

## Alternatives

Instead of introducing `Clickable`, we could have:
1. Extended or re-used the `cod-button` component. The main reason to avoid this approach is that some elements should not be considered buttons for accessibility (e.g. a screenreader should treat action buttons as links).
2. Built another standalone component. The main reason to avoid this approach is that we miss out on composability of components (one of the main pros of a UXDS). E.g. we can re-use `<cod-clickable/>` and `<cod-card/>` when building arpa buttons.

## Testing

1. Build a couple stories that render clickable cards (styled as action buttons).
2. View them in storybook.

Screenshot 2023-10-05 at 5 59 36 PM
Screenshot 2023-10-05 at 5 59 32 PM
Screenshot 2023-10-05 at 5 59 27 PM
Screenshot 2023-10-05 at 5 59 17 PM

Note: The action buttons are not equally sized in the test but can be by placing them in a card grid (like we do on the city homepage).

3. Observe cards in storybook to make sure there are no visual regressions.

Screenshot 2023-10-05 at 6 11 41 PM
Screenshot 2023-10-05 at 6 11 37 PM

4. Run test suite

```
$ yarn test-storybook
...

Test Suites: 31 passed, 31 total
Tests:       157 passed, 157 total
Snapshots:   0 total
Time:        9.096 s
Ran all test suites.
```


jedgar1mx (commented):
This might cause some ADA flags since there is no valid text or description for the link.

maxatdetroit:
Good catch @jedgar1mx , let me look at how we can better support ADA while still keeping the higher order component.

maxatdetroit:
@jedgar1mx this approach satisfies ARIA roles and other accessibility API metadata quite well. See details below and please let me know if I missed something related to a11y, and I'll go back to correct it.

<details><summary>Details</summary>
<p>


`role`, `name`, `action`, `state`, and `value` are all appropriately set for the action button example presented in the test plan. Here's a snapshot of the accessibility tree using [Firefox Accessibility Inspector](https://firefox-source-docs.mozilla.org/devtools-user/accessibility_inspector/):
```
{
    "name": "Do Something Like click on this card.",
    "role": "link",
    "actions": [
        "Jump"
    ],
    "value": "https://example.com/",
    "nodeCssSelector": ".btn",
    "nodeType": 1,
    "description": "",
    "keyboardShortcut": "",
    "childCount": 1,
    "indexInParent": 0,
    "states": [
        "focusable",
        "linked",
        "traversed",
        "opaque",
        "enabled",
        "sensitive"
    ]
}
```

Screenshot 2023-10-16 at 12 36 30 PM


</p>
</details> 

Note: The icons used in the action button are missing titles, but that should be address in a separate PR (a fix put up to set `alt` for `<cod-icon>` SVGs).

maxatdetroit:
@jedgar1mx merge conflicts are resolved here and ready for review.

jedgar1mx (commented):
Is `CardIconContainer` a necessary component? It seems that we could just add the Icon on the `CardHeader` or `CardBody` as just part of the mark up.

maxatdetroit:
Probably not necessary to have a separate component for `CardIconContainer`. It was originally intended to encapsulate the styles, but after a few iterations, the styles went down to one line (the top margin above the icon) so let me go ahead and inline the icon as markup in the header/body.

jedgar1mx (approved):

Issue #134, opened 2023-12-14
Storybook works different from production build


maxatdetroit:
**Describe the bug**

When using `<cod-table>` with `data-stacked="true"` attribute set, the design system component works different on storybook vs. production package. Specifically, the component does not properly pass the `table-stacked` class down the element stack on production.

**To Reproduce**

1. Check branch `demo.issue.134`
2. Open `src/html/demo-page-defer.html` in browser
3. Open http://localhost:6006/?path=/story/components-organisms-table--stacked-inline in browser
5. Resize both to mobile width
6. Observe they behave differently
7. Also observe the lifecycle events logged are in different order / frequency

https://github.com/CityOfDetroit/COD-Design-System/assets/143553259/e01cc843-999e-47a8-bce5-9673a0d3a8db

[storybook-console-export-2023-12-14_18-32-48.txt](https://github.com/CityOfDetroit/COD-Design-System/files/13679263/storybook-console-export-2023-12-14_18-32-48.txt)

https://github.com/CityOfDetroit/COD-Design-System/assets/143553259/3a976737-7e54-47f2-8517-576d3f36d8d5

[demo-console-export-2023-12-14_18-32-48.txt](https://github.com/CityOfDetroit/COD-Design-System/files/13679261/demo-console-export-2023-12-14_18-32-48.txt)

[detroitmi-console-export-2023-12-14_18-32-48.txt](https://github.com/CityOfDetroit/COD-Design-System/files/13679262/detroitmi-console-export-2023-12-14_18-32-48.txt)

**Expected behavior**

The table component should work similarly to the styling in storybook. Additionally, lifecycle events should be triggered in the same order / frequency between storybook and production.

maxatdetroit:
### Root Cause

The custom elements `cod-table`, `cod-table-body`, and `cod-table-row` are added to the DOM in the following order:
1. `cod-table`
2. `cod-table-body`
3. `cod-table-row`

Those same elements are defined in `index.js` in the following order:
1. `cod-table-body`
2. `cod-table-row`
3. `cod-table`

This difference in ordering is inconsequential when the script `index.js` is loaded before the DOM is parsed because the order of events is as follows:
1. Load the script: define all custom elements (but don't execute constructors, connectedCallbacks, etc. because elements are not yet declared in the DOM).
[...]

PR #140, opened 2024-01-16
Mixins for handling observed attributes & CSS classes

### LINKED ISSUE #127: Implement/adopt reactive properties and events
**Is your feature request related to a problem? Please describe.**

As the design system becomes more complex and component composition becomes more common, handling state changes within and across components requires managing element attributes and lifecycle events. Web Components provide primitives for these interactions (`attributeChangedCallback`, `observedAttributes`) but without additional structure and features those primitives can get messy pretty quickly. 

**Describe the solution you'd like**

Let's take some inspiration from LitElements and implement something similar to [Reactive Properties](https://lit.dev/docs/components/properties/) and [Events](https://lit.dev/docs/components/events/). These features should be placed in a base class `CodElement`.

Work necessary:

- [ ] Implement [Reactive Properties](https://lit.dev/docs/components/properties/) in `CodElement` base class
- [ ] Implement [Events](https://lit.dev/docs/components/events/) in `CodElement` base class
- [ ] Migrate existing components onto `CodElement` using Reactive Properties and Events when possible
- [ ] Audit how components communicate with each other to ensure they [follow the model of the DOM](https://lit.dev/docs/composition/component-composition/#passing-data-up-and-down-the-tree) – properties down and events up.

**Describe alternatives you've considered**

Alternatives:
* Adopt LitElements (or other JS/TS framework – React, Angular). I suggest we reject this approach so we can avoid external dependencies. Also, Web Components are bound to gain traction since they are browser built-in.

maxatdetroit:
## Context

As discussed in depth in #134 - [specifically this comment](https://github.com/CityOfDetroit/COD-Design-System/issues/134#issuecomment-1871446638) - we cannot rely on custom element constructors being called in the order in which they are parsed in the DOM tree. It's possible custom elements are upgraded in an order that is different than the DOM tree, and the constructors would follow suit.

The consequence of this observation is that any time we want to pass a CSS class down the DOM tree of custom elements, we must do so with [observed attributes](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#responding_to_attribute_changes) as well as `slotchange` events.

## This PR

1. Add a [mixin](https://en.wikipedia.org/wiki/Mixin) to handle observed "class attribute" changes. Here we're defining a "class attribute" as an attribute that controls the inclusion of a class on an HTMLElement, since this seems to be a common pattern on our web components. This "class attribute" pattern is used often because [Shadow DOM Encapsulation](https://javascript.info/shadow-dom#encapsulation) prevents us from using complex CSS selector descendant combinators, so the class must be explicitly passed down from the parent element.
2. Add a mixin to handle general observed attribute changes.
3. Add a utility to propagate stacked table attributes to child elements used during `slotchange` events.
4. Replace previous attribute and class propagation implementation with these new mixins and utility for the `<cod-table>` component.
5. Update the import order of components in `index.js` to import them in the opposite order they're using in the DOM (since import order shouldn't matter).

## Alternatives

I laid out a couple alternatives in CityOfDetroit/COD-Design-System#127 but they were non-starters.

A couple other alternatives that aren't necessarily non-starters:

1. Continue writing one-off implementations for observed attribute change handlers. This approach might be fine if the observed attribute has some unique handler behavior, but as we see in this PR, there are a couple main types of observed attributes ("class attributes" being one of those) that share a common pattern of implementation.
2. Use [custom states and custom state pseudo-class CSS selectors](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#custom_states_and_custom_state_pseudo-class_css_selectors) for managing the 'stacked' styles. Semantically, custom states represent state that can be modified with user interaction. It feels like a bit of a stretch in this case since user interaction isn't going to change whether the table is stacked. Also, custom internal state doesn't solve the problem of propagating that state down the DOM tree where it's needed for styling.

## Testing

1. Build the `main.js` bundle using Webpack: `yarn build`
2. Use this branch as a composer package ([see wiki](https://intranet.detroitmi.gov/node/2926)) and then test the changes on a Drupal test site with the package on this branch in use:

test-table
test-table-mobile


src/shared/js/table-stacked-mixin.js
```diff
@@ -0,0 +1,19 @@
+const tableStackedMixin = {
+  /**
+   * Handle attribute propagation for a stacked table.
+   *  codElement.handleTableStacked(
+   *    this, // the custom element.
+   *    node, // some other element lower in the DOM tree.
+   *  )
+   */
+  handleTableStacked(currentHtmlElement, targetHtmlElement) {
```
maxatdetroit: Note: We can likely remove this mixin entirely since state propagation is also handled in `observedAttributeMixin.handleObservedClassAttribute` and [observed attribute handlers are called after the attribute is initialized, when the element's declaration is parsed for the first time](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#responding_to_attribute_changes) if the element's HTML declaration includes an observed attribute.

I'll need to test this though since it's also possible the element is not defined until later (after the element declaration is parsed) during element upgrade and it's not clear what happens in that case.

jedgar1mx (changes requested):
Having an observer utility will definitely help with the nested component issues. Probably would be better to focus on attribute observation rather and classes since that is how the single components are currently managed.

maxatdetroit:
> Having an observer utility will definitely help with the nested component issues. Probably would be better to focus on attribute observation rather and classes since that is how the single components are currently managed.

I think both are going to be valuable: 
1. a utility to manage general attribute observations and 
2. a utility to specifically manage attribute observations that solely control / propagate a class

(2) is what is handled in this PR. Propagating classes in that way is the only way to select a component and it's children for styling in nested components (due to shadow DOM encapsulation). So it's fairly common in our nested components with nontrivial styling.

For (1), I'm thinking something like this in a follow-up PR:
```js
const observedAttributeMixin = {

  // ... other code here for (2)

  // code for (1)
  handleObservedAttribute(
    oldValue,
    newValue,
    onChangeCallback = (oldValue, newValue) => null,
  ) {
    onChangeCallback(oldValue, newValue);
  },
};

export { observedAttributeMixin as default };
```

Then a component would take a more declarative approach to registering the callbacks, like this:

```js
  static observedAttributeCbs = {
    'disabled': (oldValue, newValue) => {
      if (newValue is not null) {
        // query DOM for input element and disable the input.
      } else {
        // query DOM for input element and enable the input.
      }
    },
  };
  static observedAttributes = Object.keys(this.observedAttributeCbs);

  attributeChangedCallback(name, oldValue, newValue) {
    if (name in TableBody.observedAttributeCbs) {
      this.handleObservedAttribute(
        oldValue,
        newValue,
        TableBody.observedAttributeCbs[name],
      );
    }
  }
``` 

But I think we need both (1) and (2) because (2) does something more specific and if we tried to accomplish it with (1) we'd be duplicating a similar pattern over and over.

maxatdetroit:
cc @jedgar1mx for the most recent comment

jedgar1mx:
Yes, that makes sense. Let's go ahead with both observers.

maxatdetroit:
SG, I'll clean up this PR and add the general observed attribute observer to this PR too.

maxatdetroit:
Next steps (notes for self):
1. `tableStackedMixin` into a utility function in the `utilities.js` file (no need for it to be a mixin)
2. create general observer similar to pseudo-code above for observed attributes
3. add usage of `observedAttributeMixin` into `Table.js` to guard against unexpected ordering of component declarations


More context in [this comment](https://github.com/CityOfDetroit/COD-Design-System/issues/134#issuecomment-1960382017).

maxatdetroit:
@jedgar1mx the changes in this PR likely resolve the bug @sreidthomas is seeing in CityOfDetroit/detroitmi#1316. After we merge this in, I'll build the 1.0.16 release.

jedgar1mx (approved):

Issue #119, opened 2023-11-27
Allow easier use of bootstrap customization for external style libraries


maxatdetroit:
**Is your feature request related to a problem? Please describe.**

When we want to build a new bootstrap-based theme that leverages the City's design guidelines, we must copy-paste bootstrap customization from [`themed-bootstrap.scss`](https://github.com/CityOfDetroit/COD-Design-System/blob/d47e7e6897e44cbfbec38e318deee27742933103/src/scss/themed-bootstrap.scss#L24-L41) and paste it into the bootstrap SCSS initialization of the newly built theme. For example, see [code in PR #1190](https://github.com/CityOfDetroit/detroitmi/pull/1190/files#diff-0c08b670d22f0809d812dd7460a62534f6f2b9a8107a8c64096d99a19b53673fR20-R38).

**Describe the solution you'd like**

We should define our bootstrap variable customization in a [SASS mixin](https://sass-lang.com/documentation/at-rules/mixin/) that can be re-used by external libraries importing this design system.

That way, external libraries that would like to customize bootstrap with the City's design guidelines only need to do the following in their bootstrap variables SASS source file:
```
@include cod-uxds-bootstrap-variables;
```

**Describe alternatives you've considered**

Alternatives:
* A dedicated repo for the City's design guidelines based on bootstrap (call it: cod-bootstrap)
  * Pros: doesn't couple external libraries with the COD UXDS, reduces code duplication
  * Cons: requires keeping COD UXDS and cod-bootstrap in sync
* Current status quo: copy-paste bootstrap customization across 5+ themes.
  * Pros: ?
  * Cons: duplicate stylesheet code across themes and repos.

maxatdetroit:
I'm going to go ahead pick this back up. 

We're moving toward [using named slots](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_templates_and_slots) more commonly. Slots are content in the Light DOM (not shadow DOM) so they break out of our [Shadow DOM CSS encapsulation](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM#encapsulation_from_css) and use the global CSS for styling. 

In the case of the design system's storybook, the global CSS is a (mostly) non-customized bootstrap framework that differs from the City of Detroit style guide. Technically this is fine, because when the components get used, for example on detroitmi.gov, the global CSS in the Drupal theme will style those slotted elements correctly. However, this makes testing in Storybook challenging.

[...]

PR #167, opened 2024-02-27
Create component for New Action Button

sreidthomas:
### **FOR REVIEW PURPOSES ONLY, NOT READY FOR MERGE**

Creating a new action button for the re-design of detroitmi.gov

Building a new component in the design system for the new action button.

The button should accept the following attributes:

1. **icon:** the name of the icon to be used on the button

2. **primary-text:** the text to be used in the primary style (i.e. the bold text)

3. **secondary-text:** the text to be used in the secondary style (i.e. the thin text)

### **IMAGES:**

The new buttons should look like:

307073373-40898ce1-debc-41b2-b0ff-ec3daacd039c

307073237-251da56c-5a97-429b-938b-0d3108d3b138

Oftentimes, the buttons will be laid out in a grid so we have to make sure that layout looks good and the buttons are consistently sized and primary text is aligned:

303102866-1717a301-21b1-43bd-ab09-85b17b537910

### DIFF
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.css
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.css
@@ -34,6 +34,7 @@ div {
   --abutton-spacer-y: 1em;
   --abutton-spacer-x: 1em;
   --abutton-title-spacer-y: 0.5em;
+  border-bottom: solid;
 }
 
 .btn {
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.css
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.css
@@ -1,0 +1,42 @@
+div {
+  font-family: var(--font-family);
+}
+
+.action-button-container,
+.action-button-container > a,
+.action-button-container > a > .abutton {
+  width: 100%;
+  height: 100%;
+}
+
+.top-icon {
+  padding-top: 1em;
+}
+
+.abutton-title {
+  margin-bottom: var(--abutton-title-spacer-y);
+  font-weight: 900;
+  text-transform: uppercase;
+}
+
+.abutton-text:last-child {
+  margin-bottom: 0;
+}
+
+.abutton-body {
+  padding: var(--abutton-spacer-y) var(--abutton-spacer-x);
+}
+
+.abutton {
+  display: flex;
+  flex-direction: column;
+  min-width: 0;
+  --abutton-spacer-y: 1em;
+  --abutton-spacer-x: 1em;
+  --abutton-title-spacer-y: 0.5em;
+}
+
+.btn {
+  --bs-btn-padding-x: 0rem;
+  --bs-btn-padding-y: 0rem;
+}
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.css
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.css
@@ -15,12 +15,14 @@ div {
 
 .abutton-body {
   padding: var(--abutton-spacer-y) var(--abutton-spacer-x);
+  text-align: left;
 }
 
 .abutton {
   display: flex;
-  flex-direction: column;
+  flex-direction: row;
   min-width: 0;
+  padding: 1em;
   --abutton-spacer-y: 1em;
   --abutton-spacer-x: 1em;
   --abutton-title-spacer-y: 0.5em;
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.css
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.css
@@ -15,8 +15,6 @@ div {
 
 .abutton-title {
   margin-bottom: var(--abutton-title-spacer-y);
-  font-weight: 900;
-  text-transform: uppercase;
 }
 
 .abutton-text:last-child {
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.css
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.css
@@ -13,6 +13,16 @@ div {
   padding-top: 1em;
 }
 
+.abutton-title {
+  margin-bottom:var(--abutton-title-spacer-y);
+  font-weight: 900;
+  text-transform: uppercase;
+}
+
+.abutton-text:last-child {
+  margin-bottom: 0;
+}
+
 .abutton-body {
   padding: var(--abutton-spacer-y) var(--abutton-spacer-x);
   text-align: left;
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.css
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.css
@@ -13,16 +13,6 @@ div {
   padding-top: 1em;
 }
 
-.abutton-title {
-  margin-bottom: var(--abutton-title-spacer-y);
-  font-weight: 900;
-  text-transform: uppercase;
-}
-
-.abutton-text:last-child {
-  margin-bottom: 0;
-}
-
 .abutton-body {
   padding: var(--abutton-spacer-y) var(--abutton-spacer-x);
 }
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.css
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.css
@@ -14,7 +14,7 @@ div {
 }
 
 .abutton-title {
-  margin-bottom:var(--abutton-title-spacer-y);
+  margin-bottom: var(--abutton-title-spacer-y);
   font-weight: 900;
   text-transform: uppercase;
 }
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.css
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.css
@@ -7,6 +7,9 @@ div {
 .action-button-container > a > .abutton {
   width: 100%;
   height: 100%;
+  border-top: none;
+  border-right: none;
+  border-left: none;
 }
 
 .top-icon {
@@ -34,9 +37,6 @@ div {
   --abutton-spacer-y: 1em;
   --abutton-spacer-x: 1em;
   --abutton-title-spacer-y: 0.5em;
-  border-top: none;
-  border-right: none;
-  border-left: none;
 }
 
 .btn {
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.css
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.css
@@ -34,7 +34,9 @@ div {
   --abutton-spacer-y: 1em;
   --abutton-spacer-x: 1em;
   --abutton-title-spacer-y: 0.5em;
-  border-bottom: solid;
+  border-top: none;
+  border-right: none;
+  border-left: none;
 }
 
 .btn {
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.js
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.js
@@ -12,8 +12,9 @@ template.innerHTML = `
         <cod-icon data-icon="" data-size="x-large">
         </cod-icon>
       </div>
-      <div class="abutton-body">  
-        <slot></slot>
+      <div class="abutton-body">
+        <slot class="abutton-title" name="title"></slot>
+        <slot name="body"></slot>
       </div>
     </div>
   </a>
@@ -21,12 +22,14 @@ template.innerHTML = `
 `;
 
 class ActionButtonV2 extends HTMLElement {
-  static observedAttributes = ['icon'];
-
   constructor() {
+    // Always call super first in constructor
     super();
+    // Create a shadow root
     const shadow = this.attachShadow({ mode: 'open' });
     shadow.appendChild(template.content.cloneNode(true));
+
+    // Add styles
     const bootStyles = document.createElement('style');
     bootStyles.textContent = bootstrapStyles;
     const variableStyles = document.createElement('style');
@@ -39,10 +42,12 @@ class ActionButtonV2 extends HTMLElement {
   }
 
   connectedCallback() {
-    const icon = this.getAttribute('icon') || 'default-icon';
+    // Update the icon.
+    const icon = this.getAttribute('icon');
     const iconElt = this.shadowRoot.querySelector('cod-icon');
     iconElt.setAttribute('data-icon', icon);
 
+    // Update the button link and style.
     const btnColor = this.getAttribute('btn-color') ?? 'btn-outline-primary';
     const href = this.getAttribute('href');
     const target = this.getAttribute('target');
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.js
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.js
@@ -14,6 +14,8 @@ template.innerHTML = `
       </div>
       <div class="abutton-body">  
         <slot></slot>
+      <slot name="primary"></slot>
+      <slot name="secondary"></slot> 
       </div>
     </div>
   </a>
--- a/src/components/atoms/ActionButtonV2/ActionButtonV2.js
+++ b/src/components/atoms/ActionButtonV2/ActionButtonV2.js
@@ -8,7 +8,7 @@ template.innerHTML = `
 <div class="action-button-container">
   <a class="btn" role="button" href="">
     <div class="abutton">
-      <div class="w-100 top-icon">
+      <div class="top-icon">
         <cod-icon data-icon="" data-size="x-large">
         </cod-icon>
       </div>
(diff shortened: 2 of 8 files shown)


sreidthomas:
So the right now arguments are showing but I have to fill them in otherwise the story shows as undefined like in the bottom picture. It's not automatically doing it like before.

Also, I only updated the ActionButtonV2.stories.js file for now. Trying to go in order to understand better @maxatdetroit 

Screenshot 2024-02-27 at 3 33 34 PM

### **Shows as undefined when you initially click on the story:**

Screenshot 2024-02-27 at 3 31 16 PM


UPDATE:

Figured out why it was undefined:
Screenshot 2024-02-28 at 10 31 16 AM

sreidthomas:
What's the command for running prettier again @maxatdetroit?

sreidthomas:
TASK LEFT TO DO:

- [x] Change the layout of the button (CSS / Bootstrap Changes)
Got the icon to the left of the text. 
Left-aligned the text
Left-aligned and Top-Aligned the button (Tried to add rules to do this but they provided no change to the button. So I removed them."
Added some padding to the button (Will need to increase this once the drop shadow is added to the icon)

- [ ] Add the drop shadow using CSS to mimic new icons

- [ ] Figure out how to add font-weight to _primary-text_  (Switch back to using named slots and use a class for the primary-text to give it the custom CSS stylings ---> use title and body instead of primary-text and secondary-text)

- [x] Fix Prettier Issue

sreidthomas:
You'll have to check but I think I got it working. I may have set the slots up wrong though @maxatdetroit 

Screenshot 2024-02-28 at 6 55 04 PM

NOTE TO SELF:

yarn lint & yarn prettier -w "FILENAME" ==> how to fix prettier issue

src/components/atoms/ActionButtonV2/ActionButtonV2.js
```diff
@@ -0,0 +1,60 @@
...
+  <a class="btn" role="button" href="">
+    <div class="abutton">
+      <div class="top-icon">
+        <cod-icon data-icon="" data-size="x-large">
+        </cod-icon>
+      </div>
+      <div class="abutton-body">  
+        <slot></slot>
+      <slot name="primary"></slot>
+      <slot name="secondary"></slot>
```
maxatdetroit: Glad you got it working @sreidthomas 👌 

We shouldn't need to use named slots here though. From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_templates_and_slots#adding_flexibility_with_slots):

> Note: An unnamed [<slot>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) will be filled with all of the custom element's top-level child nodes that do not have the [slot](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#slot) attribute. This includes text nodes.

So we should be able to remove the extra two slots with `name=primary` and `name=secondary` and just leave the unnamed slot, then remove the slot attribute from the storybook elements. They should all get placed in the unnamed slot by default.

maxatdetroit: Also, I'll send the notes on the icon implementation this afternoon.

sreidthomas:
I am trying to add the bold/strong attribute to the primary-text but I am using slots for both primary-text and secondary-text.

Screenshot 2024-02-29 at 12 28 47 PM


 In the original, there was an h4 used for the title and it was targeted with custom CSS. How would I do this with slots?

Screenshot 2024-02-29 at 12 28 01 PM

Screenshot 2024-02-29 at 12 28 13 PM

sreidthomas:
I think I got it back with title and body as the slots but for some reason I don't think it's properly responding to the font-weight. The uppercase is working though @maxatdetroit 

Screenshot 2024-02-29 at 6 02 54 PM

### **ORIGINAL ACTIONBUTTON:**
Screenshot 2024-02-29 at 6 05 19 PM

sreidthomas:
### **Adding circle beneath the icon**

1. Edit the icon component located in _**COD-Design-System/src/components/atoms/icon/Icon.js**_

- [x] Add a boolean attribute named _'highlight'_ (This adds the circle if it is present)

2. Implement the circle

**Approach 1:** Use positioning and z-index

- [ ] Add the ellipse image as an element in the HTML
- [ ] Use relative position & z-index to position the ellipse beneath the icon

**Approach 2:** Using background-image

- [x] Leave icon HTML element as is then use CSS to create a background-image  property containing the ellipse 
- [x] Add a container around the icon(because the image is way larger than the icon) -> may need another inner container like a div so it gets displayed as a block element
- [x] Use background position to make sure it gets positioned correctly beneath the icon

maxatdetroit:
@sreidthomas , you can ignore the issue with the overridden font-weight in this PR. I wrote up some [notes here](https://github.com/CityOfDetroit/COD-Design-System/issues/119#issuecomment-1979270738) on why it's happening if you'd like to learn more, but the proper fix is outside the scope of this PR.

Regardless, the content in the action button slots will get styled correctly when used on detroitmi.gov due to the stylesheets there so this PR is safe to move forward without worrying about it. In fact, we can remove the font-weight and margin-bottom style for the title. Let's leave the uppercase style.

sreidthomas:
Screenshot 2024-03-12 at 2 38 10 PM
Screenshot 2024-03-12 at 2 38 01 PM

sreidthomas:
@maxatdetroit 
SMALL:

Screenshot 2024-03-12 at 3 37 09 PM

MEDIUM:

Screenshot 2024-03-12 at 3 37 14 PM

LARGE: 

Screenshot 2024-03-12 at 3 37 19 PM

X-LARGE:
Screenshot 2024-03-12 at 3 37 24 PM

src/components/atoms/Icon/Icon.js
```diff
@@ -1,19 +1,43 @@
...
+        itemStyles.textContent = styles;
+        shadow.appendChild(bootStyles);
+        shadow.appendChild(variableStyles);
+        shadow.appendChild(itemStyles);
   }
 
   connectedCallback() {
+    // Boolean Attribute adds circle if present
+    const isHighlighted = this.hasAttribute('is-highlighted');
+
```
maxatdetroit: Line 29 checks if all this code has already been executed and prevents the icon from being recreated if so (for performance and presentation reasons).

So this line should definitely be below line 31. It'd probably be best placed on line 68. It's a good practice to collocate variable definitions with where the variable is used.

src/components/atoms/Icon/Icon.js
```diff
@@ -34,9 +58,22 @@ export default class Icon extends HTMLElement {
...
-    const iconContainer = document.createElement('span');
-    iconContainer.innerHTML = this.getIcon(icon, size);
-    this.shadowRoot.appendChild(iconContainer);
+
+    const iconElement = document.createElement('span');
+    iconElement.innerHTML = this.getIcon(icon, size);
+    this.shadowRoot.appendChild(iconElement);
+
+    // Append the icon element to the container
+    container.appendChild(iconElement);
```
maxatdetroit: Here we're appending the same icon to two different locations in the DOM. That latter (line 67) will overwrite the first, so there's no need to have line 64 here.

src/components/atoms/Icon/circle.svg
maxatdetroit, line 1: We can remove this file from the PR since we're using the URIencoded CSS rule URL instead.

src/components/atoms/Icon/Icon.css
```diff
@@ -0,0 +1,8 @@
+/* Highlighted styles */
+.icon-container.highlighted {
+  background-image: url('data:image/svg+xml,%3Csvg%20viewBox%3D%220%200%2049%2019%22%20fill%3D%22none%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%3E%3Cellipse%20cx%3D%2224.5%22%20cy%3D%229.5%22%20rx%3D%2224.5%22%20ry%3D%229.5%22%20fill%3D%22%239FD5B3%22%2F%3E%3C%2Fsvg%3E');
+  background-repeat: no-repeat;
+  background-position-y: bottom;
+  background-size: 100% auto;
+  padding-bottom: 5px;
```
maxatdetroit: 312214742-327132d6-9c4c-42cc-97ec-acbd89eff6fc

You'll notice that the smaller icons start lurching away from the ellipse. The reason is likely due to this padding being a constant pixel value instead of a relative value like a percentage that will scale with the container size.

I would try playing around with percentages for the bottom padding. Maybe start with 10% since the XL variant looks pretty good and that is 5px bottom padding with a 54 px height icon (5 / 54 = approx. 10%).

maxatdetroit (changes requested):
I didn't review the action button element, just the icon element. Once we get that in good shape, I'll review the full PR. Left some suggestions about the icon though.

src/components/atoms/Icon/Icon.css
sreidthomas: 10% was too much too. I did .3% and it looks pretty good

Screenshot 2024-03-13 at 11 43 07 AM

src/components/atoms/Icon/Icon.js
```diff
@@ -34,9 +54,23 @@
       default:
         break;
     }
-    const iconContainer = document.createElement('span');
-    iconContainer.innerHTML = this.getIcon(icon, size);
-    this.shadowRoot.appendChild(iconContainer);
+
+    const iconElement = document.createElement('span');
+    iconElement.innerHTML = this.getIcon(icon, size);
```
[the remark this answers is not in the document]
sreidthomas: This code has been there. I never changed anything with it but its saying that its opening us to a XSS vulnerability @maxatdetroit

src/components/atoms/ActionButtonV2/ActionButtonV2.css
```diff
@@ -0,0 +1,44 @@
...
+  height: 100%;
+}
+
+.top-icon {
+  padding-top: 1em;
+}
+
+.abutton-title {
+  margin-bottom: var(--abutton-title-spacer-y);
+  font-weight: 900;
```
maxatdetroit, line 18: We might as well remove this because it has no effect on the slotted element. We can let the light DOM style the element.

src/components/atoms/ActionButtonV2/ActionButtonV2.css
```diff
@@ -0,0 +1,44 @@
...
+  text-transform: uppercase;
```
maxatdetroit: We might as well remove these and let the light DOM style the element.

src/stories/actionbuttonV2.stories.js
```diff
@@ -0,0 +1,159 @@
...
+      icon="house"
+      href="https://example.com"
+      target="_blank"
+    >
+      <h4 slot="title">Do Something</h4>
+      <p slot="body">
+        Anything can go inside an action button but it's best to keep to simple
+        text.
+      </p>
+      <img
```
maxatdetroit: Would you include the image in the slot? That will let us test if it's handling more than just text in a reasonable way.

maxatdetroit (commented):
We still have to add the highlight ellipse to the action button V2:

https://65a994a97db54752e77a065c-sujgkmkifq.chromatic.com/?path=/docs/components-atoms-actionbuttonv2--docs

Also, the v2 action button should only have a border on the bottom of the button.

308327711-66a0bddb-bfde-435c-9e46-5dbe62a459b7

sreidthomas:
> We still have to add the highlight ellipse to the action button V2:
> 
> https://65a994a97db54752e77a065c-sujgkmkifq.chromatic.com/?path=/docs/components-atoms-actionbuttonv2--docs
> 
> Also, the v2 action button should only have a border on the bottom of the button.
> 
> 308327711-66a0bddb-bfde-435c-9e46-5dbe62a459b7

Do I do this the same way I added the boolean to the Icon.js file?

src/components/atoms/ActionButtonV2/ActionButtonV2.js
```diff
@@ -0,0 +1,63 @@
...
+import bootstrapStyles from '!!raw-loader!../../../shared/themed-bootstrap.css';
+
+const template = document.createElement('template');
+
+template.innerHTML = `
+<div class="action-button-container">
+  <a class="btn" role="button" href="">
+    <div class="abutton">
+      <div class="top-icon">
+        <cod-icon data-icon="" data-size="x-large">
```
maxatdetroit: It's a little different. Unlike `Icon.js`, there's no reason to expose `is-highlighted` as an attribute on `cod-action-button-v2`. Instead, we can just statically set the attribute on the `cod-icon` in the template here:

`<cod-icon data-icon="" data-size="x-large" is-highlighted>`

sreidthomas:
Screenshot 2024-03-14 at 1 26 15 PM
Screenshot 2024-03-14 at 1 26 27 PM
Screenshot 2024-03-14 at 1 26 44 PM

@maxatdetroit I believe I fixed all changes

maxatdetroit:
@sreidthomas we missed one:

The v2 action button should only have a border on the bottom of the button.

312599072-c9039dcc-6c28-465c-8d09-6125d8385110

Also, would you try getting the icon just a little higher off the ellipse to match the mockups more closely as well as add some left and right padding so the icon is more surrounded on the left and right by the ellipse?

sreidthomas:
> @sreidthomas we missed one:
> 
> The v2 action button should only have a border on the bottom of the button.
> 
> 312599072-c9039dcc-6c28-465c-8d09-6125d8385110
> 
> Also, would you try getting the icon just a little higher off the ellipse to match the mockups more closely as well as add some left and right padding so the icon is more surrounded on the left and right by the ellipse?

How do you mean "should only have a border on the bottom of the button" @maxatdetroit ?

sreidthomas:
> @sreidthomas we missed one:
> 
> The v2 action button should only have a border on the bottom of the button.
> 
> 312599072-c9039dcc-6c28-465c-8d09-6125d8385110
> 
> Also, would you try getting the icon just a little higher off the ellipse to match the mockups more closely as well as add some left and right padding so the icon is more surrounded on the left and right by the ellipse?

Also the changes I make in the Icon.css should affect the v2 Action button right? Or do I need to make adjustments in the v2 css @maxatdetroit ?

maxatdetroit:
> How do you mean "should only have a border on the bottom of the button" @maxatdetroit ?

312973019-be4d172c-26f6-4d86-8a42-bbf9ab6c4bd7

See how the mockup (above) has only a single line on the bottom of the button? That's the bottom border. The current implementation (below) has a line on all four sides of the button. That's a top, left, right, and bottom border. We want to update the button to just have the bottom border. The [border properties](https://developer.mozilla.org/en-US/docs/Web/CSS/border-left) for each side can be controlled independently in CSS.

312939794-2473b1a7-b958-4d1f-a40d-7ea22754fafe

maxatdetroit:
> Also the changes I make in the Icon.css should affect the v2 Action button right? Or do I need to make adjustments in the v2 css @maxatdetroit ?

The `Icon.css` gets included in `<cod-icon>` component. So, since the `<cod-icon>` component is used inside `<cod-action-button-v2>` the `Icon.css` will affect the action button. Just like how the highlight ellipse works in the action button v2 even though the CSS is in `Icon.css`.

sreidthomas:
I added a border-bottom of solid. Not sure if you wanted something else @maxatdetroit. But I pushed what I have
Screenshot 2024-03-15 at 1 27 48 PM

maxatdetroit:
@sreidthomas we probably don't need to make the border solid since that's a pretty heavy line weight. We just need to remove the left, right, and top borders.

sreidthomas:
> @sreidthomas we probably don't need to make the border solid since that's a pretty heavy line weight. We just need to remove the left, right, and top borders.

Updated @maxatdetroit

maxatdetroit:
@sreidthomas I don't believe the last update worked the way we expected. The left, right, and top borders are still present on the buttons. I'm looking at the component in chromatic for this branch: https://65a994a97db54752e77a065c-btjpibzkid.chromatic.com/?path=/docs/components-atoms-actionbuttonv2--docs

maxatdetroit (changes requested):
This should be good to go once we remove the top, right, and left borders from the button.

sreidthomas:
> This should be good to go once we remove the top, right, and left borders from the button.

Is there another way to override the borders? I did: @maxatdetroit 

  border-top: none;
  border-right: none;
  border-left: none;

sreidthomas:
We're all good @maxatdetroit

maxatdetroit (approved):
giphy 

Looks great! Very nice work!

@jedgar1mx will have to review and merge.

jedgar1mx:
From the look of the original designs, it looks like the icon is bigger and position at the top instead of the center. The current size probably work better for the origin action button.

maxatdetroit:
> From the look of the original designs, it looks like the icon is bigger and position at the top instead of the center. The current size probably work better for the origin action button.

@jedgar1mx I agree but I was thinking we could merge this in then work on that styling adjustment as a followup.

Issue #202, opened 2024-04-19
Define attribute naming conventions in contributing docs


maxatdetroit:
## Is your feature request related to a problem? Please describe.

Many of our components re-declare and rename [global attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) or [standard element attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) with custom names. For example `data-disable` is intended to behave like the [`disabled`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) HTML attribute.

Also, many of our components use the `data-*` custom attribute naming convention. As discussed below, this is not necessary and superfluous.

## Describe the solution you'd like

Go through each of our components, and:
1. If the attribute behaves like a [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) or [standard element attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes)
    1. Remove `data-` prefixes (e.g. `data-id` becomes `id`)
    2. Use existing global or standard element attribute names where possible (e.g. `data-disable` becomes `disabled`)
2. Else if the attribute has a `data-` prefix
    1. Remove `data-` prefixes (e.g. `data-extra-classes` becomes `extra-classes`)
3. Else:
    1. Look for other places in the codebase where we use a similar/same attribute and make sure the attribute names and accepted values are consistent. E.g.:
        * `cod-button` has a size attribute that expects a value of `sm`, `md`, `lg` but `cod-icon` has a size attribute that expects a value of `small`, `medium`, `large` etc. They should both accept the abbreviated version.
        * `cod-action-button` has a `button-color` attribute and `cod-button` has a `data-background-color` attribute but they effectively do the same thing. They should both be named `color` and accept values from bootstrap colors (primary, secondary, etc.)
        * `cod-container` has a `data-background-color` attribute that expects `color-1`, `color-2` etc. instead of `primary`, `secondary`, etc. It should be updated to use bootstrap colors (primary, secondary, etc.).

Each time we update component attributes, we'll need to:
1. Update the attribute in the component JS implementation (e.g. CodButton.js)
3. Update the attribute definition and storybook param names in the story file (e.g. cod-button.stories.js)

## Additional Context

[...]

PR #210, opened 2024-04-30
Use common story arguments

maxatdetroit:
## Part of #202

This is another refactor to reuse storybook arguments across stories. This will make it easier for us to refactor attributes to give them consistent names and values in #202.

This is a no-op. Storybook should work the same in chromatic.

We'll merge this into dev, then merge dev into `2.0.0-alpha1` to make the uxds 2.0 easier to develop.


jedgar1mx:
It looks like some items don't have the same options to follow specially on the `sizes` and `colors`. We might need to add the missing options or just create different parameter option for those cases. For example the spinner size is not working with all the options.

maxatdetroit:
@jedgar1mx the differing size arguments was a bug. Some components don't support lg or xl. I fixed that issue.

The differing color arguments was just some missing configurations. AFAIK all the components that gained color options in this PR support those options. E.g. `<cod-spinner data-background-color="accent-secondary"/>` works OK. See the demo here: https://is.gd/JN4pid

Should be ready for review & merge.

jedgar1mx (approved):

Issue #201, opened 2024-04-19
Document best practices in contributing docs


maxatdetroit:
## 1. HTML Standard Microsyntaxes

We have some components that handle attribute values in non-standard ways. 

For example, our `<cod-button>` component has `data-disable` which expects boolean value of `true` or `false` but should really conform to the HTML standard of [boolean attributes](https://html.spec.whatwg.org/dev/common-microsyntaxes.html#boolean-attributes). 

**Describe the solution you'd like**

Generally speaking, our component APIs should adhere to all [HTML standard microsyntaxes](https://html.spec.whatwg.org/dev/common-microsyntaxes.html). Specifically, the following microstandards are relevant to our components:
* [boolean attributes](https://html.spec.whatwg.org/dev/common-microsyntaxes.html#boolean-attributes)
* [colors](https://html.spec.whatwg.org/dev/common-microsyntaxes.html#colours)
    * Note: We can also allow colors to be specified using the bootstrap color theme name (e.g. `success`, `danger`, `primary`, `secondary`, etc.)
* [space separated tokens](https://html.spec.whatwg.org/dev/common-microsyntaxes.html#space-separated-tokens)
* [comma separated tokens](https://html.spec.whatwg.org/dev/common-microsyntaxes.html#comma-separated-tokens)

Also, unless the attribute is a boolean attribute, the component should treat `attribute="undefined"` or `attribute=""` the same as omitting the attribute entirely. In Storybook stories, we should prefer omitting the attribute entirely if the storybook control is not set.

## 2. Attribute Naming Conventions

Many of our components re-declare and rename [global attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) or [standard element attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) with custom names. For example `data-disable` is intended to behave like the [`disabled`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) HTML attribute.

Also, many of our components use the `data-*` custom attribute naming convention. As discussed below, this is not necessary and superfluous.

**Describe the solution you'd like**

For each of our components:
1. If the attribute behaves like a [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) or [standard element attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes)
    1. Remove `data-` prefixes (e.g. `data-id` becomes `id`)
[...]

PR #217, opened 2024-05-20
[Button.js] Update API for HTML standard, simplify attribute naming, & migrate to named slots

sreidthomas:
### **WHAT WAS DONE:**

Update component API to adhere to HTML standard microsyntaxes, simplify attribute naming conventions, and migrate to named slots with basic markup where possible

**ISSUE:**

We have some components that handle attribute values in non-standard ways, use non-standard attribute naming conventions, and unnecessarily nest custom components instead of using named slots.

### **OVERALL TASK / JOB DUTIES:**

1. Update component APIs to adhere to all HTML standard microsyntaxes, including boolean attributes, colors, space-separated tokens, and comma-separated tokens.
2. Simplify attribute naming conventions by removing data-* prefixes and using existing standard or global attribute names where possible.
3. Migrate to named slots with basic markup where possible, removing the nesting of custom web components and the use of content passed via attributes in favor of named slots.

**ADDITIONAL NOTES:**

Nested, custom components should only be used when:
1. The nested custom component is a standalone component itself. E.g., it's OK to use <cod-button> in a custom component slot.
2. The nested content requires more than one level of depth in the DOM. Slots cannot support this case.

Each time we update component attributes or migrate to named slots, we'll need to:

• Update the attribute in the component JS implementation.
• Update the attribute definition and Storybook param names in the story file.

Components are distributed in this spreadsheet: 

[https://is.gd/ETzUCF](https://urldefense.com/v3/__https:/is.gd/ETzUCF__;!!NcXZSU8rfchoEksI!e5hJ5ZSwX0_cRP6nd1EirCNbqhYR1NAAWm6gQqMMkITmjilT-vSf0Vcw5bUdCTGB82LZF8FPkGkGpdfhNEvySJMCnzwrglp_QG4CfDk$)

### **Breakdown of individual tasks for each issue:**

**Task 201: Update component API to adhere to HTML standard microsyntaxes**

1. Identify components with attributes that do not conform to HTML standard microsyntaxes.
2. Update attribute values to adhere to standard microsyntaxes:
• For boolean attributes, ensure values are true or false instead of non-standard values.
• For color attributes, ensure values are valid color representations (e.g., hex, RGB, or color names).
• For attributes with space-separated tokens, ensure tokens are separated by spaces.
• For attributes with comma-separated tokens, ensure tokens are separated by commas.

**Task 202: Simplify attribute naming conventions**

1. Identify components with attributes that use non-standard naming conventions or unnecessary prefixes (e.g., data-*).
2. Standardize attribute names by:
• Removing data-* prefixes for attributes that behave like global or standard element attributes.
• Using existing global or standard element attribute names where possible.
• Removing unnecessary prefixes for custom attributes.
• Ensuring consistency in attribute names across components.
3. For each updated attribute or migration to named slots:
• Update the attribute in the component JS implementation.
• Update the attribute definition and Storybook param names in the story file.

**Task 205: Migrate to named slots with basic markup where possible**

1. Identify components that accept content via attributes or use nested custom components unnecessarily.
2. Migrate to named slots by:
• Removing nested custom components and using named slots instead.
• Refactoring components to accept content via named slots instead of attributes.
• Updating component documentation and examples to reflect the use of named slots.

### DIFF
--- a/src/components/atoms/Alert/Alert.js
+++ b/src/components/atoms/Alert/Alert.js
@@ -44,15 +44,15 @@ export default class Alert extends HTMLElement {
 
     // alert attributes
 
-    const icon = this.getAttribute('data-icon');
+    const icon = this.getAttribute('icon');
 
-    const iconOrder = this.getAttribute('data-icon-order');
+    const iconOrder = this.getAttribute('icon-order');
 
-    const iconSize = this.getAttribute('data-icon-size');
+    const iconSize = this.getAttribute('icon-size');
 
-    const backgroundColor = this.getAttribute('data-background-color');
+    const backgroundColor = this.getAttribute('background-color');
 
-    const extraClasses = this.getAttribute('data-extra-classes');
+    const extraClasses = this.getAttribute('extra-classes');
     this.alert.role = 'alert';
     let iconClass = '';
 
--- a/src/components/atoms/Alert/Alert.js
+++ b/src/components/atoms/Alert/Alert.js
@@ -44,15 +44,15 @@ export default class Alert extends HTMLElement {
 
     // alert attributes
 
-    const icon = this.getAttribute('icon');
+    const icon = this.getAttribute('data-icon');
 
-    const iconOrder = this.getAttribute('icon-order');
+    const iconOrder = this.getAttribute('data-icon-order');
 
-    const iconSize = this.getAttribute('icon-size');
+    const iconSize = this.getAttribute('data-icon-size');
 
-    const backgroundColor = this.getAttribute('background-color');
+    const backgroundColor = this.getAttribute('data-background-color');
 
-    const extraClasses = this.getAttribute('extra-classes');
+    const extraClasses = this.getAttribute('data-extra-classes');
     this.alert.role = 'alert';
     let iconClass = '';
 
--- a/src/components/atoms/Button/Button.js
+++ b/src/components/atoms/Button/Button.js
@@ -51,7 +51,9 @@ export default class Button extends HTMLElement {
     imgAlt
       ? btnClasses.push('cod-button--img')
       : btnClasses.push('cod-button--not-img');
-    close === 'true' ? btnClasses.push('btn-close') : 0;
+    if (close) {
+      btnClasses.push('btn-close');
+    }
     btn.className = btnClasses.join(' ');
 
     if (icon) {
--- a/src/components/atoms/Button/Button.js
+++ b/src/components/atoms/Button/Button.js
@@ -12,8 +12,8 @@ export default class Button extends HTMLElement {
 
   connectedCallback() {
     // Button attributes
-    const close = this.getAttribute('data-close');
-    const link = this.getAttribute('data-link');
+    const close = this.getAttribute('close');
+    const link = this.getAttribute('link');
     const btnID = this.getAttribute('data-id');
     const ariaLabel = this.getAttribute('data-aria-label');
     const primary = this.getAttribute('data-primary');
--- a/src/components/atoms/Button/Button.js
+++ b/src/components/atoms/Button/Button.js
@@ -12,11 +12,11 @@ export default class Button extends HTMLElement {
 
   connectedCallback() {
     // Button attributes
-    const close = this.getAttribute('close');
+    const close = this.hasAttribute('close');
     const link = this.getAttribute('link');
     const btnID = this.getAttribute('id');
     const ariaLabel = this.getAttribute('aria-label');
-    const primary = this.getAttribute('primary');
+    const primary = this.hasAttribute('primary');
     const backgroundColor = this.getAttribute('background-color');
     const shape = this.getAttribute('shape');
     const icon = this.getAttribute('icon');
@@ -28,7 +28,7 @@ export default class Button extends HTMLElement {
     const size = this.getAttribute('size');
     const extraClasses = this.getAttribute('extra-classes');
     const label = this.getAttribute('label');
-    const disableStatus = this.getAttribute('disabled');
+    const disableStatus = this.hasAttribute('disabled');
     // Building Button component
     const btn = document.createElement('button');
     const btnClasses = ['btn'];
--- a/src/components/atoms/Button/Button.js
+++ b/src/components/atoms/Button/Button.js
@@ -16,18 +16,18 @@ export default class Button extends HTMLElement {
     const link = this.getAttribute('link');
     const btnID = this.getAttribute('id');
     const ariaLabel = this.getAttribute('aria-label');
-    const primary = this.getAttribute('data-primary');
-    const backgroundColor = this.getAttribute('data-background-color');
-    const shape = this.getAttribute('data-shape');
-    const icon = this.getAttribute('data-icon');
-    const iconSize = this.getAttribute('data-icon-size');
-    const iconOrder = this.getAttribute('data-icon-order');
-    const hiddenLabel = this.getAttribute('data-hidden-label');
-    const imgSrc = this.getAttribute('data-img');
-    const imgAlt = this.getAttribute('data-img-alt');
-    const size = this.getAttribute('data-size');
-    const extraClasses = this.getAttribute('data-extra-classes');
-    const label = this.getAttribute('data-label');
+    const primary = this.getAttribute('primary');
+    const backgroundColor = this.getAttribute('background-color');
+    const shape = this.getAttribute('shape');
+    const icon = this.getAttribute('icon');
+    const iconSize = this.getAttribute('icon-size');
+    const iconOrder = this.getAttribute('icon-order');
+    const hiddenLabel = this.getAttribute('hidden-label');
+    const imgSrc = this.getAttribute('img');
+    const imgAlt = this.getAttribute('img-alt');
+    const size = this.getAttribute('size');
+    const extraClasses = this.getAttribute('extra-classes');
+    const label = this.getAttribute('label');
     const disableStatus = this.getAttribute('data-disable');
     // Building Button component
     const btn = document.createElement('button');
@@ -58,8 +58,8 @@ export default class Button extends HTMLElement {
       // Loading icon
       const iconContainer = document.createElement('span');
       const activeIcon = document.createElement('cod-icon');
-      activeIcon.setAttribute('data-icon', icon);
-      activeIcon.setAttribute('data-size', iconSize);
+      activeIcon.setAttribute('icon', icon);
+      activeIcon.setAttribute('size', iconSize);
       iconContainer.appendChild(activeIcon);
       btn.innerText = label;
       if (iconOrder === 'left') {
--- a/src/components/atoms/Button/Button.js
+++ b/src/components/atoms/Button/Button.js
@@ -28,7 +28,7 @@ export default class Button extends HTMLElement {
     const size = this.getAttribute('size');
     const extraClasses = this.getAttribute('extra-classes');
     const label = this.getAttribute('label');
-    const disableStatus = this.getAttribute('data-disable');
+    const disableStatus = this.getAttribute('disabled');
     // Building Button component
     const btn = document.createElement('button');
     const btnClasses = ['btn'];
--- a/src/components/atoms/Button/Button.js
+++ b/src/components/atoms/Button/Button.js
@@ -38,9 +38,9 @@ export default class Button extends HTMLElement {
     }
     disableStatus === 'true' ? (btn.disabled = true) : (btn.disabled = false);
     btn.setAttribute('aria-label', `${ariaLabel || ''}`);
-    if (primary === 'true') {
+    if (primary) {
       btnClasses.push(`btn-${backgroundColor}`);
-    } else if (primary === 'false') {
+    } else {
       btnClasses.push(`btn-outline-${backgroundColor}`);
     }
     shape === 'square'
--- a/src/components/atoms/Button/Button.js
+++ b/src/components/atoms/Button/Button.js
@@ -58,8 +58,8 @@ export default class Button extends HTMLElement {
       // Loading icon
       const iconContainer = document.createElement('span');
       const activeIcon = document.createElement('cod-icon');
-      activeIcon.setAttribute('icon', icon);
-      activeIcon.setAttribute('size', iconSize);
+      activeIcon.setAttribute('data-icon', icon);
+      activeIcon.setAttribute('data-size', iconSize);
       iconContainer.appendChild(activeIcon);
       btn.innerText = label;
       if (iconOrder === 'left') {
--- a/src/components/atoms/Button/Button.js
+++ b/src/components/atoms/Button/Button.js
@@ -14,8 +14,8 @@ export default class Button extends HTMLElement {
     // Button attributes
     const close = this.getAttribute('close');
     const link = this.getAttribute('link');
-    const btnID = this.getAttribute('data-id');
-    const ariaLabel = this.getAttribute('data-aria-label');
+    const btnID = this.getAttribute('id');
+    const ariaLabel = this.getAttribute('aria-label');
     const primary = this.getAttribute('data-primary');
     const backgroundColor = this.getAttribute('data-background-color');
     const shape = this.getAttribute('data-shape');
--- a/src/components/atoms/ModalFooter/ModalFooter.js
+++ b/src/components/atoms/ModalFooter/ModalFooter.js
@@ -43,7 +43,7 @@ export default class ModalFooter extends HTMLElement {
 
     const btnExtraClasses = this.getAttribute('data-button-extra-classes');
 
-    const extraClasses = this.getAttribute('extra-classes');
+    const extraClasses = this.getAttribute('data-extra-classes');
 
     const modalFooterClasses = ['modal-footer'];
     this.closeBtn.setAttribute('img-alt', '');
--- a/src/components/atoms/ModalFooter/ModalFooter.js
+++ b/src/components/atoms/ModalFooter/ModalFooter.js
@@ -43,12 +43,12 @@ export default class ModalFooter extends HTMLElement {
 
     const btnExtraClasses = this.getAttribute('data-button-extra-classes');
 
-    const extraClasses = this.getAttribute('data-extra-classes');
+    const extraClasses = this.getAttribute('extra-classes');
 
     const modalFooterClasses = ['modal-footer'];
-    this.closeBtn.setAttribute('data-img-alt', '');
-    this.closeBtn.setAttribute('data-icon', '');
-    this.closeBtn.setAttribute('data-label', 'Close');
+    this.closeBtn.setAttribute('img-alt', '');
+    this.closeBtn.setAttribute('icon', '');
+    this.closeBtn.setAttribute('label', 'Close');
     this.closeBtn.setAttribute('data-bs-dismiss', 'modal');
 
     // TODO: Fix old ESLint errors - see issue #1099
--- a/src/components/atoms/ModalHeader/ModalHeader.js
+++ b/src/components/atoms/ModalHeader/ModalHeader.js
@@ -47,15 +47,15 @@ export default class ModalHeader extends HTMLElement {
 
     const btnDark = this.getAttribute('data-button-dark');
 
-    const extraClasses = this.getAttribute('data-extra-classes');
+    const extraClasses = this.getAttribute('extra-classes');
 
     const modalHeaderClasses = ['modal-header'];
     this.modalTitle.className = 'modal-title';
     this.modalTitle.id = `${parentID}-label`;
-    this.closeBtn.setAttribute('data-img-alt', '');
-    this.closeBtn.setAttribute('data-icon', '');
-    this.closeBtn.setAttribute('data-close', 'true');
-    this.closeBtn.setAttribute('data-bs-dismiss', 'modal');
+    this.closeBtn.setAttribute('img-alt', '');
+    this.closeBtn.setAttribute('icon', '');
+    this.closeBtn.setAttribute('close', 'true');
+    this.closeBtn.setAttribute('bs-dismiss', 'modal');
 
     // TODO: Fix old ESLint errors - see issue #1099
     // eslint-disable-next-line eqeqeq
@@ -66,7 +66,7 @@ export default class ModalHeader extends HTMLElement {
     // TODO: Fix old ESLint errors - see issue #1099
     // eslint-disable-next-line eqeqeq
     btnDark == 'true'
-      ? this.closeBtn.setAttribute('data-extra-classes', 'btn-close-white')
+      ? this.closeBtn.setAttribute('extra-classes', 'btn-close-white')
       : 0;
     this.modalHeader.className = modalHeaderClasses.join(' ');
     this.closeBtn.addEventListener('click', this._onClick);
--- a/src/components/atoms/ModalHeader/ModalHeader.js
+++ b/src/components/atoms/ModalHeader/ModalHeader.js
@@ -47,7 +47,7 @@ export default class ModalHeader extends HTMLElement {
 
     const btnDark = this.getAttribute('data-button-dark');
 
-    const extraClasses = this.getAttribute('extra-classes');
+    const extraClasses = this.getAttribute('data-extra-classes');
 
     const modalHeaderClasses = ['modal-header'];
     this.modalTitle.className = 'modal-title';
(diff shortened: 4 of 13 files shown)


sreidthomas:
### **UPDATE AS OF MAY 20, 2024:**

This component is used in the OffcanvasHeader.js file so in order to properly update OffcanvasHeader.js so attributes need to be changed here:

`this.closeBtn.setAttribute('data-img-alt', '');`
`this.closeBtn.setAttribute('data-icon', '');`
`this.closeBtn.setAttribute('data-close', 'true');`
`this.closeBtn.setAttribute('data-bs-dismiss', parentID);`

**In Progress:**
### **Task 202: Simplify attribute naming conventions part 1 & 2**

Did more research on standard, global and custom attributes. Here is what I have so far:

- [x] data-close: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-link: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-id: Standard attribute (global attribute) **(Updated per instruction --> Using existing global or standard element attribute names where possible. [https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id](url))**

- [x] data-aria-label: Standard attribute (global attribute) **(Updated per instruction --> Using existing global or standard element attribute names where possible. [https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes](url)   )**

- [x] data-primary: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-background-color: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-shape: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-icon: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-icon-size: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-icon-order: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-hidden-label: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-img: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-img-alt: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-size: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-extra-classes: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-label: Custom attribute **(Updated per instruction --> Standardize attribute names by:  Removing unnecessary prefixes for custom attributes.)**

- [x] data-disable: Standard attribute (global attribute) **(Updated per instruction --> Using existing global or standard element attribute names where possible.  [https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled](url))**

###  **Task 202: Simplify attribute naming conventions part 3**

For all the attributes updated in Button.js, the same attributes and params were updated in the Button.stories.js file per instruction --> **For each updated attribute or migration to named slots:
• Update the attribute in the component JS implementation.
• Update the attribute definition and Storybook param names in the story file.**

sreidthomas:
### **QUESTIONS**: @maxatdetroit 

1. If the component is not using slots are we still supposed to migrate to named slots?
2. When you migrate to named slots are we only updating the template and story files? Is there always something specific that needs to be updated with each element besides the template and story files e.g. EventListeners?
3. When we migrate to named templates do we change how we add styles, varStyles, & bootstrapStyles to the component?
4. Are we only checking for the tasks with issue 201 in the .js files or the stories.js file too?

sreidthomas:
### **CURRENT ISSUES AS OF MAY 21, 2024:**
### **DO THIS ONCE YOU GET TO COD-ICON PR**
The cod-button uses cod-icon so the attributes in cod-icon need to be updated to properly reflect the changes in cod-button

NEXT COMPONENT TO START ON --> cod-icon

- [ ] Update the attributes in cod-icon and re-test to make sure cod-button is working correctly with the icons

sreidthomas:
### **NEW TASK FOR MAY 21, 2024:**

### Task 201: Update component API to adhere to HTML standard microsyntaxes

Identify components with attributes that do not conform to HTML standard microsyntaxes. Update attribute values to adhere to standard microsyntaxes:

- [x] For boolean attributes, ensure values are true or false instead of non-standard values.

- [x] For color attributes, ensure values are valid color representations (e.g., hex, RGB, or color names).

- [x]  For attributes with space-separated tokens, ensure tokens are separated by spaces.

- [x]  For attributes with comma-separated tokens, ensure tokens are separated by commas.

- There are no attributes in Button.js file that use comma-separated tokens

maxatdetroit:
@sreidthomas 

**Question 1:**

>     If the component is not using slots are we still supposed to migrate to named slots?

No. If the component isn't using slots, then there is nothing to migrate from.


**Question 2:**

>     When you migrate to named slots are we only updating the template and story files? Is there always something specific that needs to be updated with each element besides the template and story files e.g. EventListeners?

Whenever you modify a component's public API (i.e. attribute name, attribute behavior, or component slots), we have to look for all the other places in the design system where that component is used. E.g. for `cod-button` do a code search for `cod-button` and see where the button is used in the design system. Anywhere it's used, we have to update which attributes and slots are used to match what we updated (e.g. if `cod-modal` is using `cod-button` it may still be using the `data-close` attribute which should be changed to just `close`).


**Question 3:**

>     When we migrate to named templates do we change how we add styles, varStyles, & bootstrapStyles to the component?

Do you mean named slots? If so, it depends. If the styling changes unexpectedly when we move the content into a slot, then yes, we might have to update the styles to target the elements correctly since they've moved in the component layout.


**Question 4:**

>     Are we only checking for the tasks with issue 201 in the .js files or the stories.js file too?


I think the answer to question 2 should answer this question as well but let me know if that's not clear.

sreidthomas:
### **There are no slots in the Button.js file so Task 205: Migrate to named slots with basic markup where possible is not needed.**

sreidthomas:
NOTE TO SELF:

yarn lint & yarn prettier -w "FILENAME" ==> how to fix prettier issue

sreidthomas:
Can you review this one for me and see if I'm on the right track? I think I have to make changes to the cod-icon in order for this to work correctly but I don't want to start on another component yet @maxatdetroit

src/components/atoms/Button/Button.js
```diff
@@ -58,8 +58,8 @@ export default class Button extends HTMLElement {
       // Loading icon
       const iconContainer = document.createElement('span');
       const activeIcon = document.createElement('cod-icon');
-      activeIcon.setAttribute('data-icon', icon);
-      activeIcon.setAttribute('data-size', iconSize);
+      activeIcon.setAttribute('icon', icon);
+      activeIcon.setAttribute('size', iconSize);
```
maxatdetroit: You could leave these as `data-icon` and `data-size` in this PR and handle updating the `cod-icon` later. These attributes are for `cod-icon` so you can leave them as is and perform the updates to each component independently (in separate PRs).

maxatdetroit (commented):
Yep, definitely on the right track.

> I think I have to make changes to the cod-icon in order for this to work correctly

That's not necessarily true. You could leave the `cod-icon` the same and handle it in a separate PR. That's what I'd recommend, because otherwise this PR will get pretty big. I left a comment about what I mean.

However, you will have to change all the places where `cod-button` gets used across the design system to use the correct attribute names that you changed. Here's a list of all the places where the button is used:
1. Places where it's created via HTML markup: https://github.com/search?q=repo%3ACityOfDetroit%2FCOD-Design-System+%3Ccod-button&type=code
2. Places where it's created via JS DOM API: https://github.com/search?q=repo%3ACityOfDetroit%2FCOD-Design-System+document.createElement%28%27cod-button%27&type=code

sreidthomas:
> Yep, definitely on the right track.
> 
> > I think I have to make changes to the cod-icon in order for this to work correctly
> 
> That's not necessarily true. You could leave the `cod-icon` the same and handle it in a separate PR. That's what I'd recommend, because otherwise this PR will get pretty big. I left a comment about what I mean.
> 
> However, you will have to change all the places where `cod-button` gets used across the design system to use the correct attribute names that you changed. Here's a list of all the places where the button is used:
> 
> 1. Places where it's created via HTML markup: https://github.com/search?q=repo%3ACityOfDetroit%2FCOD-Design-System+%3Ccod-button&type=code
> 2. Places where it's created via JS DOM API: https://github.com/search?q=repo%3ACityOfDetroit%2FCOD-Design-System+document.createElement%28%27cod-button%27&type=code

Yes that's what I meant starting another PR for it. But I see that I should keep moving with any components that are using the cod-button

src/components/atoms/Button/Button.js
sreidthomas: Question: if I change these in cod-icon (separate PR) I still have to come back and make this change later right @maxatdetroit?

sreidthomas:
> Yep, definitely on the right track.
> 
> > I think I have to make changes to the cod-icon in order for this to work correctly
> 
> That's not necessarily true. You could leave the `cod-icon` the same and handle it in a separate PR. That's what I'd recommend, because otherwise this PR will get pretty big. I left a comment about what I mean.
> 
> However, you will have to change all the places where `cod-button` gets used across the design system to use the correct attribute names that you changed. Here's a list of all the places where the button is used:
> 
> 1. Places where it's created via HTML markup: https://github.com/search?q=repo%3ACityOfDetroit%2FCOD-Design-System+%3Ccod-button&type=code
> 2. Places where it's created via JS DOM API: https://github.com/search?q=repo%3ACityOfDetroit%2FCOD-Design-System+document.createElement%28%27cod-button%27&type=code

You want me to change the places that are using the old attributes within this same PR @maxatdetroit?

maxatdetroit:
> Yes that's what I meant starting another PR for it. But I see that I should keep moving with any components that are using the cod-button

@sreidthomas ahh gotchya. Thanks for clarifying.

> You want me to change the places that are using the old attributes within this same PR @maxatdetroit?

Yea, let's go ahead and do it in one PR. If we tried to do it in a separate PR, then it'd be difficult to merge the PRs in a safe way (that doesn't break the design system).

src/components/atoms/Button/Button.js
```diff
@@ -12,23 +12,23 @@ export default class Button extends HTMLElement {
...
+    const icon = this.getAttribute('icon');
+    const iconSize = this.getAttribute('icon-size');
+    const iconOrder = this.getAttribute('icon-order');
+    const hiddenLabel = this.getAttribute('hidden-label');
+    const imgSrc = this.getAttribute('img');
+    const imgAlt = this.getAttribute('img-alt');
+    const size = this.getAttribute('size');
+    const extraClasses = this.getAttribute('extra-classes');
+    const label = this.getAttribute('label');
+    const disableStatus = this.getAttribute('disabled');
```
maxatdetroit: Heads up @sreidthomas : 

#201 talks about 'boolean attributes'. `close`, `primary`, and `disabled` should all be boolean attributes. That means we have to also do the following to make those attributes behave like boolean attributes:
1. update how `Button.js` gets/handles those attributes
2. update how various components and stories set those attributes

src/stories/card.stories.js
```diff
@@ -35,18 +35,18 @@ export const Basic = () => html`
...
-        data-img=""
-        data-img-alt=""
-        data-icon=""
-        data-icon-order=""
-        data-icon-size=""
-        data-shape="fluid"
-        data-aria-label=""
-        data-link="https://detroitmi.gov"
+        primary="true"
+        disable="false"
```
maxatdetroit: E.g. since these are boolean attributes, they should be like this now:

```html
<cod-button
  primary
  <!-- Notice how we don't even set `disabled` here since its omission should be treated like 'false'. -->
>
```

src/stories/card.stories.js
```diff
@@ -35,18 +35,18 @@ export const Basic = () => html`
...
+        label="Link"
+        background-color="primary"
+        img=""
+        img-alt=""
+        icon=""
+        icon-order=""
+        icon-size=""
```
maxatdetroit: Also, there's a comment in #201 like this:

> Also, unless the attribute is a boolean attribute, the component should treat attribute="undefined" or attribute="" the same as omitting the attribute entirely. In Storybook stories, we should prefer omitting the attribute entirely if the storybook control is not set.

So for these non-boolean attributes where we're passing `""` or `"undefined"`, let's remove them entirely instead of continuing to pass empty values.

Note: We may need to also update `Button.js` to handle these attributes differently.

maxatdetroit (changes requested):
@sreidthomas here's a couple more comments regarding #201 that we want to take into account.

sreidthomas:
Places where it's created via JS DOM API: https://github.com/search?q=repo%3ACityOfDetroit%2FCOD-Design-System+document.createElement%28%27cod-button%27&type=code

The cod-button is just created in these places, there are no attributes so there's nothing that needs updating here right @maxatdetroit?

sreidthomas:
### **TO-DO AS OF JULY 15, 2024:**

- [x] Go back over issue 201 involving boolean values

maxatdetroit:
> Places where it's created via JS DOM API: https://github.com/search?q=repo%3ACityOfDetroit%2FCOD-Design-System+document.createElement%28%27cod-button%27&type=code
> 
> The cod-button is just created in these places, there are no attributes so there's nothing that needs updating here right @maxatdetroit?

That's not necessarily true. Those attributes can still be set using the JS DOM API. For example, see how the `close` attribute is set here: https://github.com/CityOfDetroit/COD-Design-System/blob/7165fee15a9c92c35b4f1efb5791524e23a6abdc/src/components/atoms/OffcanvasHeader/OffcanvasHeader.js#L52. That would need to be changed.

I will say however it seems the attributes are used less commonly in the JS DOM API inside the design system based on the search results you share. I looked at maybe 4 of the search results and only the `OffcanvasHeader` example I shared sets one of the boolean attributes via the JS DOM API.

sreidthomas:
### **CHANGES TO JAVASCRIPT TO BUTTON.JS FOR HANDLING BOOLEANS:**

**CODE PIECE 1:**
```
const primary = this.getAttribute('primary');
const close = this.getAttribute('close');
const disableStatus = this.getAttribute('disabled');
```

**CHANGE**: These should be using hasAttribute because they are dealing with Booleans: [https://developer.mozilla.org/en-US/docs/Web/API/Element/hasAttribute](url)

**CODE PIECE 2:**
`disableStatus === 'true' ? (btn.disabled = true) : (btn.disabled = false);`

**CHANGE:** This doesn't need to be a ternary but not sure what to change it to ===> **btn.disabled = disableStatus;**


**CODE PIECE 3:**
`if (primary === 'true') {
  btnClasses.push(`btn-${backgroundColor}`);
} else if (primary === 'false') {
  btnClasses.push(`btn-outline-${backgroundColor}`);
}`

**CHANGE:** Working on it ==> this doesn't need to be compared to true or false since we know its a boolean and being treated as such **(remove the === true / false since we are already treating primary as a boolean)**

`if (primary) {
  btnClasses.push(`btn-${backgroundColor}`);
} else {
  btnClasses.push(`btn-outline-${backgroundColor}`);
}`

**CODE PIECE 4:**

`close === 'true' ? btnClasses.push('btn-close') : 0;`

**CHANGE:** Working on it ==> **close is already a boolean so no need to set it equal to true. can remove ternary as well**
```

if (close) {
  btnClasses.push('btn-close');
}
```

sreidthomas:
### **TO-DO FOR JULY 19, 2024:**

TASK 1: Update the boolean attributes in these files:

- [x] dropdown.stories.js
- [x] navbar.stories.js
- [x] form.stories.js
- [x] card.stories.js
- [x] nav.stories.js
- [x] buttongroup.stories.js

TASK 2: Remove non-boolean attributes where we're passing "" or "undefined" completely from these files:
- [x] dropdown.stories.js
- [x] navbar.stories.js
- [x] form.stories.js
- [x] card.stories.js
- [x] nav.stories.js
- [x] buttongroup.stories.js

- [x] TASK 3: Update Button.js to handle the attributes where we're passing "" or "undefined" 

- [x] TASK 4: Update these lines in components/atoms/OffcanvasHeader/OffcanvasHeader.js

  ``` 
this.closeBtn.setAttribute('data-img-alt', '');
this.closeBtn.setAttribute('data-icon', '');
this.closeBtn.setAttribute('data-close', 'true');
```

**Comment by sreidthomas** [comment]
I believe I covered everything but can you go over to make sure @maxatdetroit

File: src/stories/buttongroup.stories.js [hunk 5]
```diff
@@ -136,7 +82,7 @@ export const CheckboxButtonGroup = () => html`
     ></cod-form-check>
   </cod-button-group>
   <br /><br />
-  <cod-button-group data-type="group" data-label="basic example">
+  <cod-button-group data-type="group" label="basic example">
```
maxatdetroit: We still have to use `data-label` here since this is for the button group component which we haven't fixed yet.

src/stories/card.stories.js
```diff
@@ -35,18 +35,12 @@ export const Basic = () => html`
...
-        data-img=""
-        data-img-alt=""
-        data-icon=""
-        data-icon-order=""
-        data-icon-size=""
-        data-shape="fluid"
-        data-aria-label=""
-        data-link="https://detroitmi.gov"
+        primary
+        disable="false"
```
maxatdetroit: I think this should be `disabled` now. (with the 'd').

maxatdetroit (changes requested):
It looks like we got most of the places where the `<cod-button>` is created via HTML, but not the places where the `<cod-button>` is created via JS DOM API.

Here are the other places we have to change: https://github.com/search?q=repo%3ACityOfDetroit%2FCOD-Design-System%20createElement(%27cod-button%27)&type=code

For example, the `data-*` attribute names are still used for the buttons here: https://github.com/CityOfDetroit/COD-Design-System/blob/cfd278caa8561dce2ffbc1df8af2676726aab297/src/components/atoms/OffcanvasHeader/OffcanvasHeader.js#L50-L53

sreidthomas:
### **TO-DO:**

**Update  attributes for these components/files:**

- [x] Alert.js
- [x] button.stories.js
- [x] ModalFooter.js
- [x] Map.js
- [x] OffcanvasHeader.js
- [x] ModalHeader.js

src/components/atoms/Alert/Alert.js
```diff
@@ -44,15 +44,15 @@ export default class Alert extends HTMLElement {
...
+    const iconOrder = this.getAttribute('icon-order');
 
-    const iconSize = this.getAttribute('data-icon-size');
+    const iconSize = this.getAttribute('icon-size');
 
-    const backgroundColor = this.getAttribute('data-background-color');
+    const backgroundColor = this.getAttribute('background-color');
 
-    const extraClasses = this.getAttribute('data-extra-classes');
+    const extraClasses = this.getAttribute('extra-classes');
```
maxatdetroit: @sreidthomas this is a breaking change and shouldn't be part of this PR.

While these attributes share the same name as the attributes on the button component, these attributes are owned and controlled by the alert component. This PR is focused on updating the button component attributes, not the alert component attributes. 

The place to look in this file for button attributes would be around [line 100 - 106](https://github.com/CityOfDetroit/COD-Design-System/blob/e967a55376002709169ec7081a07d76d3887844e/src/components/atoms/Alert/Alert.js#L100-L106) where the JS DOM API is used to create a button component. If the `data-icon`, `data-icon-size` etc. were used on the button component on those lines, we would change them as part of this PR. However, those attributes are not used on the button in this file, so we can omit any changes to this file entirely.

src/components/atoms/ModalFooter/ModalFooter.js
```diff
@@ -43,12 +43,12 @@ export default class ModalFooter extends HTMLElement {
 
     const btnExtraClasses = this.getAttribute('data-button-extra-classes');
 
-    const extraClasses = this.getAttribute('data-extra-classes');
+    const extraClasses = this.getAttribute('extra-classes');
```
maxatdetroit: This is another example where we're changing how the attributes on the ModalFooter component work instead of focusing on the attributes on the button component. 

However, the attributes below (lines 49-51) are button component attributes with correct updates.

src/stories/buttongroup.stories.js
```diff
@@ -8,85 +8,31 @@ export default {
 };
 
 export const Basic = () => html`
-  <cod-button-group data-type="group" data-label="basic example">
+  <cod-button-group data-type="group" label="basic example">
```
maxatdetroit: This shouldn't be changed. The `data-label` attribute on the button group is owned and controlled by the `<cod-button-group>` component which has not been updated yet. This PR should be focused on updating attributes for the button component only.

maxatdetroit (changes requested):
I left a few comments where I noticed we are updating components incorrectly and conflating button component attributes with other components' attributes. I may have missed some places, so please take a look over the entire PR and see if I missed any similar places.

Also, please test your changes as you make them by viewing and interacting with the components you edit in storybook. I caught these issues because I tested this PR in storybook and noticed the Alert component is broken with these changes.

src/components/atoms/ModalHeader/ModalHeader.js
```diff
@@ -47,15 +47,15 @@ export default class ModalHeader extends HTMLElement {
 
     const btnDark = this.getAttribute('data-button-dark');
 
-    const extraClasses = this.getAttribute('data-extra-classes');
+    const extraClasses = this.getAttribute('extra-classes');
```
maxatdetroit: This shouldn't be changed. It's not a button attribute. This is an attribute on ModelHeader.

src/components/atoms/OffcanvasHeader/OffcanvasHeader.js
```diff
@@ -47,13 +47,13 @@ export default class OffcanvasHeader extends HTMLElement {
...
     this.offcanvasTitle.className = 'offcanvas-title';
     this.offcanvasTitle.id = `${parentID}-label`;
-    this.closeBtn.setAttribute('data-img-alt', '');
-    this.closeBtn.setAttribute('data-icon', '');
-    this.closeBtn.setAttribute('data-close', 'true');
-    this.closeBtn.setAttribute('data-bs-dismiss', parentID);
+    this.closeBtn.setAttribute('img-alt', '');
+    this.closeBtn.setAttribute('icon', '');
+    this.closeBtn.setAttribute('close', 'true');
+    this.closeBtn.setAttribute('bs-dismiss', parentID);
```
maxatdetroit: In ModalFooter we left this attribute unchanged but in ModalHeader and here we removed `data-` prefix. Which is correct?

src/components/organisms/Map/Map.js
```diff
@@ -214,16 +214,16 @@ export default class Map extends HTMLElement {
             closeMapBtn.addEventListener('click', () => {
               app[0] ? app[0].setAttribute('data-app-state', 'results') : 0;
             });
-            closeMapBtn.setAttribute('data-primary', true);
-            closeMapBtn.setAttribute('data-label', 'x');
-            closeMapBtn.setAttribute('data-size', 'large');
+            closeMapBtn.setAttribute('primary', true);
+            closeMapBtn.setAttribute('label', 'x');
+            closeMapBtn.setAttribute('size', 'large');
             closeMapBtn.setAttribute('data-hover', false);
```
maxatdetroit, line 220: Why does this attribute still have the `data-` prefix?

maxatdetroit (changes requested):
There's still a handful of places where this comment is still applicable:

> I left a few comments where I noticed we are updating components incorrectly and conflating button component attributes with other components' attributes. I may have missed some places, so please take a look over the entire PR and see if I missed any similar places.

Please take another look.

maxatdetroit (approved):
Looks good now!

Btw, don't forget to set the base branch to `2.0.0-alpha1` for the uxds 2.0 PRs. I just updated this one on github.

sreidthomas:
Going back over the code for this component and each task. For 201, we had to update everything to meet HTML Standard, and while studying, I learned that **href** is standard instead of using **link** like how we're doing. Or is there a reason why we're using link over href @maxatdetroit?

maxatdetroit:
> Going back over the code for this component and each task. For 201, we had to update everything to meet HTML Standard, and while studying, I learned that **href** is standard instead of using **link** like how we're doing. Or is there a reason why we're using link over href @maxatdetroit?

That's a **really** good catch @sreidthomas . There's no reason why we're using the `link` attr over `href`. Let's use `href` moving forward during the UXDS 2.0 migration.

PR #218, opened 2024-05-22
Cleanup.Spinner

SBBlee:
Simplifies attribute naming conventions (removes data- prefix) as mentioned in issue #202 

Capture3


SBBlee:
@maxatdetroit, I just wanted to clarify that #202 was the only issue that pertained to this component

maxatdetroit:
Yes, good clarification. That was correct when you asked, but I did notice while reviewing this PR that the storybook stories are setting missing controls to `undefined` in the code examples, so I added this request to #201:

> Also, unless the attribute is a boolean attribute, the component should treat attribute="undefined" the same as omitting the attribute entirely. In Storybook stories, we should prefer omitting the attribute entirely if the storybook control is not set.

Would you please update the Storybook story for the spinner to omit attributes if they're not set in the story (e.g. `size` and `display-type` should be omitted from the component in the earlier screenshot)? Then, let's make sure the spinner still functions as expected when those attributes are omitted.

maxatdetroit:
Also, all PRs for 201, 202, and 205 should be based on the `2.0.0-alpha1` branch. I failed to mention that the other day - sorry! You'll need to:

`git rebase 2.0.0-alpha1`

This is a case where a `rebase` is required instead of using `merge`. After you've done that and pushed the changes to this PR. We can update the base branch on the PR as well (using the edit button near the PR title):

image

maxatdetroit (changes requested):
I think we moved this implementation to https://github.com/CityOfDetroit/COD-Design-System/pull/219 due to base branch issues on this PR. Is that right? Can we close this PR in favor of https://github.com/CityOfDetroit/COD-Design-System/pull/219?

PR #219, opened 2024-05-24
Clean-up.Spinner

SBBlee:
Removes data prefix and omits unused attributes as mentioned in issue #201 and #202 

Capture


src/stories/spinner.stories.js
```diff
@@ -29,10 +29,18 @@ export default {
...
-  spinner.setAttribute('data-background-color', args.backgroundColor);
-  spinner.setAttribute('data-size', args.size);
-  spinner.setAttribute('data-display-type', args.displayType);
+  spinner.setAttribute('type', args.type);
+  spinner.setAttribute('background-color', args.backgroundColor);
+  if (args.size){
+    spinner.setAttribute('size', args.size);
+  } else {
+    delete spinner.size;
+  }
```
maxatdetroit: This shouldn't be necessary because the `size` attribute will never have been set if we reach this else condition.

src/stories/spinner.stories.js
```diff
@@ -29,10 +29,18 @@ export default {
...
+  if (args.displayType){
+    spinner.setAttribute('display-type', args.displayType);
+  } else {
+    delete spinner.displayType;
+  }
```
maxatdetroit: Similar comment regarding unnecessary delete statement.

src/stories/spinner.stories.js
maxatdetroit: Removing this else statement entirely should continue to work as expected.

maxatdetroit (approved):
Just a minor comment regarding the storybook attribute implementation. Please address then I'll merge this in.

maxatdetroit (changes requested):
Requesting changes so we can fix the lingering comments from the previous review.

maxatdetroit (approved):
Thanks for making those tweaks.

PR #251, opened 2024-07-31
Update display fonts to match headers

### LINKED ISSUE #247: Update display fonts to match headers
## The Problem

Similar to the findings in https://github.com/CityOfDetroit/detroitmi/pull/1579, we found that our bootstrap configuration has our heading elements styled differently than the bootstrap display fonts.

This makes it challenging to use display fonts or headers interchangeably when needed.

## The Solution

Update the bootstrap configuration in the design system to set the display font sizes and weights to match their heading counterparts. E.g. h1 should have the same font size and weight as display-1, h2 the same as display-2, etc.

sreidthomas:
### **Description:**
This PR introduces custom display font sizes to our _bootstrap_variables.scss file to ensure consistency with the DetroitMI style guide. The custom sizes will override the default Bootstrap heading font sizes.

### **Issue:**
The default Bootstrap heading font sizes do not align with our design requirements. We need to define custom sizes to maintain visual consistency across our project.

### **Changes:**

Defined a new map variable $display-font-sizes with custom display font sizes.
Updated the default heading font size variables ($h1-font-size, $h2-font-size, etc.) to use values from the $display-font-sizes map.

### **PICTURE OF THE DISPLAY SIZES:**

displaysizes

**LINK TO ISSUE:** [https://github.com/orgs/CityOfDetroit/projects/36/views/1?pane=issue&itemId=72099163](url)


src/scss/_bootstrap-variables.scss
```diff
@@ -27,6 +37,14 @@ $font-size-base: 1rem; // Assumes the browser default, typically `16px`
 $font-size-sm: $font-size-base;
 $font-size-lg: $font-size-base * 1.25;
 
+// Override the default headings font sizes.
+$h1-font-size: map-get($display-font-sizes, 1);
+$h2-font-size: map-get($display-font-sizes, 2);
+$h3-font-size: map-get($display-font-sizes, 3);
+$h4-font-size: map-get($display-font-sizes, 4);
+$h5-font-size: map-get($display-font-sizes, 5);
+$h6-font-size: map-get($display-font-sizes, 6);
```
maxatdetroit: This is a really smart idea to simplify the code and tie the display font and header fonts together. Nice work!

Lines 49-54 are overwriting your great work though on these variables. Would you remove those later lines?

maxatdetroit (changes requested):
One comment on the code, otherwise this is great. Would you please test your changes by adding some headers and display font styled elements inside a component then take a screenshot to confirm they are the same size and weight?

sreidthomas:
> One comment on the code, otherwise this is great. Would you please test your changes by adding some headers and display font styled elements inside a component then take a screenshot to confirm they are the same size and weight?

You want to pick just any component inside the design system? And do I save the changes? @maxatdetroit

maxatdetroit:
> You want to pick just any component inside the design system? And do I save the changes?

Any component where the headers and display fonts are inside the shadow DOM would be an OK test. For example, the [card component](https://cityofdetroit.github.io/COD-Design-System/?path=/docs/components-organisms-card--docs) has the card body inside the shadow DOM so you could create a card and test that way. No, you don't need to save the changes, just make sure the behavior is right and share a screenshot on the PR description (so others can verify how you tested and for future reference).

maxatdetroit (approved):
will-ferrel-beautiful

Issue #243, opened 2024-07-16
Fix screen reader support for dropdown


maxatdetroit:
## The Problem

The dropdown menu does not announce opening / closing to screenreaders.

See https://github.com/CityOfDetroit/detroitmi/issues/1577 for an example of this in the wild.

## To Reproduce

1. Go to https://cityofdetroit.github.io/COD-Design-System/?path=/docs/components-molecules-dropdown--docs
2. Launch a screen reader
3. Click on the dropdown
4. Observe that the dropdown opening is not announced

**Screenshots**

![image](https://github.com/user-attachments/assets/bb6b3e9b-b1de-477e-b129-ea2453319dac)

## Expected behavior

The dropdown opening is announced.

## The Solution

This is most likely due to the lack of the following attributes on the dropdown button and menu:
- `aria-controls="[menu-id]"` on the button
- `aria-labelledby="[button-id]"` on the dropdown menu
- `aria-hidden="true/false"` on the dropdown menu

To fix this, we need to update the Dropdown.js COD design system to add and manage the state of the aforementioned attributes on the `cod-button` and `cod-dropdown-menu`. When the dropdown menu is opened/closed, the state of `aria-hidden` should be updated to reflect the state of the menu.

Issue #205, opened 2024-04-24
Add contributing docs regarding named slots with basic markup


maxatdetroit:
**Is your feature request related to a problem? Please describe.**

Many of our components either:
1. accept content via attributes. E.g. `<cod-button>` accepts image, label, and icon as attributes when these pieces of content could be passed in via named slots.
2. require the use of slotted custom components. E.g. `<cod-card>` uses `<cod-card-body>` unnecessarily.

Both of these approaches have performance implications because they require custom Javascript functions to be executed that are typically slower than built-in browser methods and behaviors (like native slotting behavior). This is especially true of slotted custom components which require an entire custom element lifecycle to execute.

**Describe the solution you'd like**

* Where possible (see additional context) remove the nesting of custom web components in favor of named slots.
* Remove the use of content passed via attributes in favor of named slots. This is also a recommendation in [web.dev](https://is.gd/wEksoE).

**Additional context**

Nested, custom components should only be used when:
1. The nested custom component is a standalone component itself. E.g. it's OK to use `<cod-button>` in a custom component slot.
2. The nested content requires more then one level of depth in the DOM. [Slots cannot support this case](https://javascript.info/slots-composition).

**Work breakdown**

It'll help us avoid merge conflicts if we break down the work this way per component. Components are distributed in this spreadsheet: https://is.gd/ETzUCF.

maxatdetroit:
This is going to be handled in https://github.com/CityOfDetroit/COD-Design-System/issues/315 so this issue will be more about updating our contributing README docs to talk about these principles being used in the design system.

maxatdetroit:
Moved to #201

PR #253, opened 2024-08-06
Fix.Dropdown

SBBlee:
Adds attributes for screen readers as mentioned in issue #1577 [https://github.com/CityOfDetroit/detroitmi/issues/1577](url) and issue #243

### DIFF
--- a/src/components/molecules/Dropdown/Dropdown.js
+++ b/src/components/molecules/Dropdown/Dropdown.js
@@ -51,6 +51,9 @@ export default class Dropdown extends HTMLElement {
       this.querySelector(
         'cod-button[data-bs-toggle="dropdown"]',
       ).addEventListener('click', this._onClick);
+      this.querySelector(
+        'cod-button[data-bs-toggle="dropdown"]')
+      .setAttribute('aria-controls', '#dropdown-id');
     }
     if (!this.shadowRoot.querySelector('div')) {
       this.shadowRoot.appendChild(this.dropdown);
@@ -68,11 +71,20 @@ export default class Dropdown extends HTMLElement {
         this.parentElement
           .querySelector('cod-dropdown-menu')
           .setAttribute('data-show', 'false');
+        this.parentElement
+          .querySelector('cod-dropdown-menu') 
+          .setAttribute('aria-hidden', 'true');           
       } else {
         this.setAttribute('aria-expanded', 'true');
         this.parentElement
-          .querySelector('cod-dropdown-menu')
+          .querySelector('cod-dropdown-menu') 
           .setAttribute('data-show', 'true');
+        this.parentElement
+          .querySelector('cod-dropdown-menu') 
+          .setAttribute('aria-labelledby', '#button-id');
+        this.parentElement
+          .querySelector('cod-dropdown-menu') 
+          .setAttribute('aria-hidden', 'false');          
       }
     }
   }
--- a/src/components/molecules/Dropdown/Dropdown.js
+++ b/src/components/molecules/Dropdown/Dropdown.js
@@ -51,9 +51,14 @@ export default class Dropdown extends HTMLElement {
       this.querySelector(
         'cod-button[data-bs-toggle="dropdown"]',
       ).addEventListener('click', this._onClick);
-      this.querySelector(
-        'cod-button[data-bs-toggle="dropdown"]')
-      .setAttribute('aria-controls', '#dropdown-id');
+      const buttonElt = this.querySelector(
+        'cod-button[data-bs-toggle="dropdown"]',
+      );
+      const menuId = this.querySelector('cod-dropdown-menu')
+      .getAttribute('id');
+      buttonElt 
+      .setAttribute('aria-controls', menuId);  
+      
     }
     if (!this.shadowRoot.querySelector('div')) {
       this.shadowRoot.appendChild(this.dropdown);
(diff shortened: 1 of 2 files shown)


SBBlee:
@maxatdetroit I just submitted what I have so far. I still need help with correctly adding button-id and menu-id. Does the stories need to be re-written to achieve this?

src/components/molecules/Dropdown/Dropdown.js
```diff
@@ -68,11 +71,20 @@ export default class Dropdown extends HTMLElement {
         this.parentElement
           .querySelector('cod-dropdown-menu')
           .setAttribute('data-show', 'false');
+        this.parentElement
+          .querySelector('cod-dropdown-menu') 
+          .setAttribute('aria-hidden', 'true');
```
maxatdetroit: Just a minor comment: we can store the result of the query selector instead of performing the query twice. Like this:

```js
const dropdownMenu = this.parentElement
  .querySelector('cod-dropdown-menu');

dropdownMenu
  .setAttribute('data-show', 'false')
  .setAttribute('aria-hidden', 'true'); 
```

src/components/molecules/Dropdown/Dropdown.js
```diff
@@ -68,11 +71,20 @@ export default class Dropdown extends HTMLElement {
...
+          .setAttribute('aria-hidden', 'true');           
       } else {
         this.setAttribute('aria-expanded', 'true');
         this.parentElement
-          .querySelector('cod-dropdown-menu')
+          .querySelector('cod-dropdown-menu') 
           .setAttribute('data-show', 'true');
+        this.parentElement
+          .querySelector('cod-dropdown-menu') 
+          .setAttribute('aria-labelledby', '#button-id');
```
maxatdetroit: We'll have to get the button ID from the DOM. Like this:


```js
const buttonID = this.parentElement
  .querySelector('cod-button')
  .getAttribute('id');
```

Then we can replace `'#button-id'` with `buttonID`.

src/components/molecules/Dropdown/Dropdown.js
```diff
@@ -51,6 +51,9 @@ export default class Dropdown extends HTMLElement {
       this.querySelector(
         'cod-button[data-bs-toggle="dropdown"]',
       ).addEventListener('click', this._onClick);
+      this.querySelector(
+        'cod-button[data-bs-toggle="dropdown"]')
+      .setAttribute('aria-controls', '#dropdown-id');
```
maxatdetroit: We'll have to follow a similar approach described below on line 84 to get the dropdown ID.

src/stories/dropdown.stories.js
```diff
@@ -18,7 +18,7 @@ export const Basic = () => html`
       >
         Dropdown button
       </button>
-      <ul class="dropdown-menu">
+      <ul class="dropdown-menu" id="dropdown-menu">
```
maxatdetroit: These HTML elements with the class `no-wc` are not actually included in the custom element. They get filtered out by Javascript (see line 27-28 of Dropdown.js). The reason they exist is to support browsers that don't have Javascript enabled. 

The elements that actually get included start at line 29 are what get included in custom element shadow DOM. Those are the elements where we should set the IDs.

So, for example, line 29-38 should be:
```html
    <cod-button
      id="dropdownControl"
      data-label="Dropdown button"
      data-label="Dropdown button"
      data-background-color="primary"
      data-background-color="primary"
      data-primary="true"
      data-primary="true"
      data-img-alt=""
      data-img-alt=""
      data-icon=""
      data-icon=""
      data-bs-toggle="dropdown"
      data-bs-toggle="dropdown"
      aria-expanded="false"
      aria-expanded="false"
      data-extra-classes="dropdown-toggle"
      data-extra-classes="dropdown-toggle"
    ></cod-button>
```

maxatdetroit: Then we would want to similarly set an ID on the dropdown menu element as well.

maxatdetroit (changes requested):
@SBBlee this is looking great and moving in the right direction. Left some comments about how to set/get the button and menu IDs.

To answer your question: yes the IDs must be set in the storybook story. Then, they're fetched inside the Dropdown.js.

maxatdetroit:
@SBBlee I tested with a screen reader tool (like [Narrator for Windows](https://support.microsoft.com/en-us/windows/complete-guide-to-narrator-e4397a0d-ef4f-b386-d8ae-c172f109bdb1)) and the dropdown still isn't announcing the expanded / collapsed state.

This is likely because the `aria-*` attributes need to be placed on the `button` and `ul` elements specifically instead of the `cod-button` and `cod-dropdown-menu` elements. That will require edits to `Dropdown.js`, `DropdownMenu.js`, and `Button.js` which will be a lot of throw-away work since we're going to migrate to basic markup in the design system as part of https://github.com/CityOfDetroit/COD-Design-System/issues/205.

I think we're actually better off going with option (2) described in [the original issue](https://github.com/CityOfDetroit/detroitmi/issues/1577) which is to use basic bootstrap markup in the GDYT menu template for the dropdown instead of leveraging the design system. Do you have time to implement that change this afternoon so we can get it in for tomorrows release?

maxatdetroit (changes requested):
@SBBlee I tested with a screen reader tool (like [Narrator for Windows](https://support.microsoft.com/en-us/windows/complete-guide-to-narrator-e4397a0d-ef4f-b386-d8ae-c172f109bdb1)) and the dropdown still isn't announcing the expanded / collapsed state.

This is likely because the `aria-*` attributes need to be placed on the `button` and `ul` elements specifically instead of the `cod-button` and `cod-dropdown-menu` elements. That will require edits to `Dropdown.js`, `DropdownMenu.js`, and `Button.js` which will be a lot of throw-away work since we're going to migrate to basic markup in the design system as part of https://github.com/CityOfDetroit/COD-Design-System/issues/205.

I think we're actually better off going with option (2) described in [the original issue](https://github.com/CityOfDetroit/detroitmi/issues/1577) which is to use basic bootstrap markup in the GDYT menu template for the dropdown instead of leveraging the design system. Do you have time to implement that change this afternoon so we can get it in for tomorrows release?

SBBlee:
@maxatdetroit I tried the bootstrap dropdown first and there was an was a contradicting element stopping the implementation. I can't remember the exact element you said that caused the issue, maybe it was Popper.

maxatdetroit:
Jeez what a mess 🤦‍♂️ I apologize for the back-and-forth. We're learning a lot about the issues in our design system. 

The original issue with using the bootstrap dropdown in https://github.com/CityOfDetroit/detroitmi/pull/1603 was that the  navbar was built using the design system. I removed the use of the design system in the navbar as part of https://github.com/CityOfDetroit/detroitmi/pull/1611, so that shouldn't be an issue any longer and we can use a bootstrap dropdown.

SBBlee:
@maxatdetroit I pulled dev and still not seeing the removal of design system elements. I also merged into my branch and having the same issue as before 

Dev
image

maxatdetroit:
@SBBlee that screenshot is correct. 

> still not seeing the removal of design system elements

The design system elements that were removed were the `cod-nav`, `cod-navbar`, `cod-navbar-brand`, etc. The `cod-dropdown` and child elements should be the only design system elements left and your task would be to replace those with bootstrap components like you started in https://github.com/CityOfDetroit/detroitmi/pull/1603/files.

Or am I misunderstanding what you mean?

SBBlee:
@maxatdetroit I misunderstood, I thought the cod-dropdown was updated as well.  I was expecting to see behavior similar to your [PR #1611](https://github.com/CityOfDetroit/detroitmi/pull/1611). I believe you said the bootstrap classes were correct, but the behavior still isn't ideal.  This is what I have so far

https://github.com/user-attachments/assets/a1e82cff-04e9-45b2-b373-73596af190da


image

maxatdetroit:
@SBBlee would you push your latest changes to CityOfDetroit/detroitmi#1603 and we can continue the discussion on that PR? I think it's probably pretty close but debugging over screenshots is sort of tough.

maxatdetroit:
Closing out this PR since we're moving the work to https://github.com/CityOfDetroit/detroitmi/pull/1603

PR #280, opened 2025-01-08
GovBanner Component Creation

### LINKED ISSUE #279: Create .gov banner component
# .gov Banner Component 

The goal of this task is to create a new .gov banner component in the stable build of the design system.

## Mockups

### Mobile

Image

### Desktop

Image

### Functionality

When someone clicks the down chevron, the banner should expand and show the following content:

Image

## Requirements

  - Note: The text container on the right side of the banner should allow the text to break onto a newline if the component width is decreased enough.

sreidthomas:
# Create `.gov` Banner Component in Stable Build of the Design System

## Goal  
Implement a `.gov` banner component in the stable build of the design system that expands to reveal content when the down chevron is clicked.

---

### Mobile

Image

### Desktop

Image

### **Functionality Requirements**  

When someone clicks the down chevron, the banner should expand and show the following content:

Image

https://github.com/user-attachments/assets/1a4a39c1-b690-40c9-afe2-93ce78276f31

### **TEST RESULTS FOR GOVBANNER:**
image
image


src/stable/components/organisms/GovBanner/GovBanner.css
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  .city-name {
+    margin-top: 0.25rem;
+    flex-shrink: 0;
+  }
+}
+
+/* Mobile-specific overrides */
+@media (max-width: 550px) {
```
maxatdetroit: Media queries are a good example of when using SCSS will make our lives easier. Will you migrate this to SCSS and leverage [bootstrap breakpoint media queries](https://getbootstrap.com/docs/5.3/layout/breakpoints/#media-queries)? The script you wrote in this repo will need to be run in order to generate the associated CSS.

src/stable/components/organisms/GovBanner/GovBanner.js
```diff
@@ -0,0 +1,91 @@
...
+    </div>
+  </header>
+  <div id="content" class="content-container" hidden>
+    <div class="info-section">
+      <div class="info-item">
+        <div class="icon-circle">
+          <span class="gov-icon">🏛️</span>
+        </div>
+        <div>
+          <h3>Official websites use .gov</h3>
```
maxatdetroit: Header elements are typically reserved for the main content of a page.

Also, if an H3 is used, an H2 and H1 must precede it. We don't want to make any assumptions about other headers on the page and their order.

Let's use a normal `span` here and style the typography to look similar to a header instead.

src/stable/stories/govbanner.stories.js
```diff
@@ -0,0 +1,45 @@
...
+              <h3>Secure .gov websites use HTTPS</h3>
+              <p>
+                A lock or https:// means you've safely connected to the .gov
+                website. Share sensitive information only on official, secure
+                websites.
+              </p>
+            </div>
+          </div>
+        </div>
+      </div>
```
maxatdetroit: Since we're removed the slots in the template, this markup is no longer used. You should be able to remove it without impacting the story.

src/stable/stories/govbanner.stories.js
```diff
@@ -0,0 +1,45 @@
+import { html } from 'lit-html';
+import '../../stable/components/organisms/GovBanner/cod-gov-banner';
+
+export default {
+  title: 'Organisms/GovBanner',
```
maxatdetroit: Let's drop this approach of bucketing components in 'Atoms/Molecules/Organisms' moving forward. Eventually we want the storybook to be used by non-technical folks (for picking out components for their pages) and those aren't very non-tech-friendly buckets.

src/stable/components/organisms/GovBanner/GovBanner.js
```diff
@@ -0,0 +1,91 @@
...
+          <h3>Secure .gov websites use HTTPS</h3>
+          <p>A <b>lock</b> (🔒) or <b>https://</b> means you've safely connected to the .gov website. Share sensitive information only on official, secure websites.</p>
+        </div>
+      </div>
+    </div>
+  </div>
+</div> 
+`;
+
+class GovBanner extends HTMLElement {
```
maxatdetroit: That thing you were doing with JS object getter accessors in an [earlier commit on this PR](https://github.com/CityOfDetroit/COD-Design-System/pull/280/commits/709b9f2548ea8ada8eea278ab1cc070e2d892ca0) was a good pattern. Would you bring it back and get it working? 

The high level idea is to keep a JS object private variable and HTML attribute in sync by using attributeChange event handlers and JS object getter/setter accessors. Then, the HTML attribute acts as the public API for the component and the variable is the internal state representation.

Here's an example with a simple dropdown:

https://gist.github.com/maxatdetroit/ab09739e847256761676dd44823bfd65

src/stable/components/organisms/GovBanner/GovBanner.js
```diff
@@ -0,0 +1,91 @@
...
+
+  _handleToggle() {
+    const content = this.shadowRoot.querySelector('#content');
+    const button = this.shadowRoot.querySelector('.chevron-container');
+    if (content && button) {
+      const isExpanded = button.getAttribute('aria-expanded') === 'true';
+      button.setAttribute('aria-expanded', !isExpanded);
+      content.hidden = isExpanded;
+    }
+  }
```
maxatdetroit: Also, notice how in the example above, we emit a custom event: https://gist.github.com/maxatdetroit/ab09739e847256761676dd44823bfd65#file-simple-dropdown-example-js-L111

This is a good practice so that other components (or people using your component) can react to your component changing state. Could we add that in?

src/stable/components/organisms/GovBanner/GovBanner.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+export { GovBanner as default };
```
maxatdetroit: Could we write a [component test](https://storybook.js.org/docs/writing-tests/component-testing#write-a-component-test) for this component? It should test the following:

When the header is clicked, its `expanded` property is changed, or its `expanded` attribute is changed, then the component does the following:
* Shows/hides the content element.
* Emits a JS event.
* Attribute & property reflect the same value as the one provided

maxatdetroit (changes requested):
This is **really** nice work. I left a few requests – some of which you were doing previously then removed – that will make this an exemplar for all future components in the design system.

src/stable/components/organisms/GovBanner/GovBanner.css
sreidthomas: If I have to use the script, wouldn't it have to be updated since it coded to deal with the experimental components before we updated the directory structure @maxatdetroit?

src/stable/components/GovBanner/GovBanner.js
```diff
@@ -0,0 +1,132 @@
...
+
+  constructor() {
+    super();
+    const shadow = this.attachShadow({ mode: 'open' });
+    shadow.appendChild(template.content.cloneNode(true));
+  }
+
+  get expanded() {
+    return (
+      this.hasAttribute('expanded') && this.getAttribute('expanded') === 'true'
```
maxatdetroit: `this.getAttribute('expanded') === 'true'` is not necessary. Check out the [HTML spec for boolean attributes](https://html.spec.whatwg.org/dev/common-microsyntaxes.html#boolean-attributes). It states that as long as the attribute is present on the element, it's considered True. It doesn't need an explicit value of 'true'.

maxatdetroit: This should also impact how you implement `set expanded(value)` and `attributeChangedCallback` methods (i.e. they shouldn't expect the `expanded` attribute to have a value of `true`).

src/stable/components/GovBanner/GovBanner.js
```diff
@@ -0,0 +1,132 @@
...
+      new CustomEvent('expandedchange', {
+        detail: { expanded: isExpanded },
+        bubbles: true,
+      }),
+    );
+  }
+
+  connectedCallback() {
+    this._setupListeners();
+    this._updateExpandedState(this.expanded);
```
maxatdetroit, line 92: `this.expanded` is not initialized. We should give it a default value of false.

src/stable/components/GovBanner/GovBanner.js
```diff
@@ -0,0 +1,132 @@
...
+      <div class="official-text">
+        An official website of the City of Detroit.
+        <span class="know-text">
+          Here's how you know
+          <button class="chevron-container" aria-expanded="false" aria-controls="content">
+            <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-chevron-down" viewBox="0 0 16 16">
+              <path fill-rule="evenodd" d="M1.646 4.646a.5.5 0 0 1 .708 0L8 10.293l5.646-5.647a.5.5 0 0 1 .708.708l-6 6a.5.5 0 0 1-.708 0l-6-6a.5.5 0 0 1 0-.708"/>
+            </svg>
+          </button>
+        </span>
```
maxatdetroit: Right now, only the chevron is clickable. We need the entire 'know-text' component to be clickable (and that element should be the button as well). Inspect the banner on this site for an example: https://www.usds.gov/

src/stable/stories/govbanner.stories.js
```diff
@@ -0,0 +1,75 @@
...
+  title: 'Stable/GovBanner',
+  tags: ['autodocs'],
+};
+
+export const Default = {
+  render: () => html` <cod-gov-banner> </cod-gov-banner> `,
+};
+
+export const ExpandedBehavior = {
+  render: () => html` <cod-gov-banner> </cod-gov-banner> `,
```
maxatdetroit, line 16: Should this story show up on the storybook docs? I just see the Default story in your video on the PR description. Or maybe the video in the description of the PR was created before we added this story? 

If this shows up in the storybook docs, then with a name like 'ExpandedBehavior', I'd expect this to start in the expanded state (to show users how to start the component expanded):
`<cod-gov-banner expanded></cod-gov-banner>`

If so, maybe we could move the test to the 'default' story.

maxatdetroit (changes requested):
Just a few comments to resolve, then this is ready to merge.

Also, I see in the PR description video that at certain screen widths, the text overflows off the component (and is not visible). Is that still happening? If so, we need to fix that so that the text wraps onto the next line on overflow instead of being hidden.

src/stable/components/GovBanner/GovBanner.js
```diff
@@ -0,0 +1,131 @@
+import styles from '!!raw-loader!./GovBanner.css';
+import varStyles from '!!raw-loader!../../../shared/variables.css';
+import bootstrapStyles from '!!raw-loader!../../../shared/themed-bootstrap.css';
+
+const template = document.createElement('template');
+template.innerHTML = `
+  <style>
+    ${bootstrapStyles}
+    ${varStyles}
```
maxatdetroit: I believe we can actually remove these stylesheets since we're not using any bootstrap utility classes. Would you try that?

maxatdetroit (approved):
Wowza - nice work! Just left one comment about trying to remove unnecessary style sheets.

maxatdetroit (changes requested):
Sorry @sreidthomas, I noticed one more thing we have to handle before merge:

The text in the banner needs to wrap on small screens instead of flowing off the end of the component.

See what I mean here: https://65a994a97db54752e77a065c-pqpytfknni.chromatic.com/?path=/story/stable-govbanner--gov-banner&globals=backgrounds.grid:!false;outline:!true

You can use storybook to test various viewport sizes easily:


https://github.com/user-attachments/assets/f174bc22-2d3e-466a-8e5d-8227cd185e95

sreidthomas:
> Sorry @sreidthomas, I noticed one more thing we have to handle before merge:
> 
> The text in the banner needs to wrap on small screens instead of flowing off the end of the component.
> 
> See what I mean here: https://65a994a97db54752e77a065c-pqpytfknni.chromatic.com/?path=/story/stable-govbanner--gov-banner&globals=backgrounds.grid:!false;outline:!true
> 
> You can use storybook to test various viewport sizes easily:
> 2025-02-17.13-18-46.mp4

@maxatdetroit 
https://github.com/user-attachments/assets/fb6b4f8c-775e-4ade-9cc1-17961325a0e0

maxatdetroit (approved):
nice work ty!

PR #282, opened 2025-01-17
SectionNavigation Component Creation

### LINKED ISSUE #277: Create section navigation component for skip links
# Section Navigation Component

The goal of this task is to create a new section navigation component in the stable build of the design system.

## Mockups

### Mobile

Image

Image

Image

### Desktop

Image

Image

## Requirements

  - When the width of the viewport is smaller than 992px (large bootstrap breakpoint), the mobile design should be used.
  - Recommendation: Try to implement the mobile <> desktop transition using stylesheet media queries. If that doesn't work, try using ResizeObserver or window.matchMedia in Javascript instead.
- The individual skip links should be placed inside the section navigation using slots and custom web components (see https://shoelace.style/components/menu as an example of what the API should look like)

sreidthomas:
### Mobile

Image

Image

Image

### Desktop

Image

Image

https://github.com/user-attachments/assets/8622bbc3-546e-49b7-8730-32b753bab37b

### **TEST RESULTS:**

image


src/stable/components/organisms/SectionNavigation/SectionNavigation.js
```diff
@@ -0,0 +1,61 @@
...
+const template = document.createElement('template');
+template.innerHTML = `
+<style>
+${bootstrapStyles}
+${varStyles}
+${styles}
+</style>
+<div class="section-container">
+  <div class="section-header">
+    <h2>On This Page</h2>
```
maxatdetroit: Let's use a slot for this instead of statically setting an `h2` element. That way the user of the component can decide what the appropriate element is here (which also helps avoid accessibility issues with out-of-order headings).

src/stable/components/organisms/SectionNavigation/SectionNavigation.js
```diff
@@ -0,0 +1,61 @@
...
+    <button type="button" class="toggle-button" aria-label="Toggle navigation" aria-expanded="false">
+      <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-chevron-down" viewBox="0 0 16 16">
+        <path fill-rule="evenodd" d="M1.646 4.646a.5.5 0 0 1 .708 0L8 10.293l5.646-5.647a.5.5 0 0 1 .708.708l-6 6a.5.5 0 0 1-.708 0l-6-6a.5.5 0 0 1 0-.708"/>
+      </svg>
+    </button>
+  </div>
+  <nav class="section-nav">
```
maxatdetroit: This `nav` element [should also be explicitly labeled](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#description) so that people with disabilities know that the section header described what these links are for.

See [example 5 here to see what I mean](https://www.w3.org/WAI/WCAG21/Techniques/html/H48).

src/stable/components/organisms/SectionNavigation/SectionNavigation.css
```diff
@@ -0,0 +1,127 @@
...
+    flex-direction: column;
+  }
+
+  ::slotted(a) {
+    display: block !important;
+    padding: 0.75rem 1rem !important;
+    margin: 0 !important;
+    border: none !important;
+    color: #000 !important;
+    text-decoration: none !important;
```
maxatdetroit: I believe you asked me (1) why these rules needs `!important` and (2) how we can avoid using `!important` (since its a bad practice). 

The answer to (1): its probably because `::slotted` pseudo selector has pretty low specificity when it comes to targeting objects in the Light DOM (like `a` elements).

The answer to (2): we should avoid styling slotted elements entirely. The slotted elements like the `a` element are part of the Light DOM, so they get styled by the stylesheets in the Light DOM (not these stylesheets here in the Shadow DOM. This is intentional: it allows us to style our layout and component without having to worry about the actual *content* looking different than the rest of the website (e.g. an `a` element in our component should look the same as an `a` element on other parts of the website).

So what do we do here? For the text styles (e.g. `color`, `text-decoration`), let's remove the `slotted` styling and let the `a` element be styled by the light DOM. For the layout styles (e.g. `display`, `padding`, etc.), let's add a container element (like `li` suggested in a separate comment) that is part of the shadow DOM and encapsulates the slotted `a`, then we'll style that element layout instead.

src/stable/components/organisms/SectionNavigation/SectionNavigation.css
```diff
@@ -0,0 +1,127 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: 0.5rem 2rem;
+}
+
+::slotted(a) {
+  color: #000 !important;
+  text-decoration: none !important;
+  border-bottom: none !important;
+  padding: 0.25rem 0 !important;
```
maxatdetroit: Ditto re: how to avoid using `!important`.

src/stable/components/organisms/SectionNavigation/SectionNavigation.js
```diff
@@ -0,0 +1,61 @@
...
+    <slot name="nav-items"></slot>
```
maxatdetroit: When we have groups of links, we need to keep them together in a HTML list for accessibility purposes. 

The same [example 5 here is a good reference](https://www.w3.org/WAI/WCAG21/Techniques/html/H48).

Two ways to approach this:
1. We could wrap the slot in a `<ul>`, and then tell the user of the component they must place a `<li><a>...</a></li>` in the slot. 
2. We could wrap the slot in a `<ul>`, then listed for a [slot change event ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/slotchange_event) on the `nav-items` slot. When there is a slot change, if the element passed to the slot is not wrapped in a `<li>` item, we wrap it in an `<li>` item for them. 

#2 is preferred here since it will guarantee there are no accessibility issues when this component is used.

src/stable/stories/sectionnavigation.stories.js
```diff
@@ -0,0 +1,21 @@
+import { html } from 'lit-html';
+import '../../stable/components/organisms/SectionNavigation/cod-section-navigation';
+
+export default {
+  title: 'Organisms/SectionNavigation',
+  tags: ['autodocs'],
+};
+
+export const Default = {
```
maxatdetroit, line 48: We should add a test to this component like we're doing for the gov banner component. I'll let you come up with the different conditions / scenarios to test. What do you think they should be?

src/stable/components/organisms/SectionNavigation/SectionNavigation.css
```diff
@@ -0,0 +1,127 @@
...
+  border: none;
+  padding: 0;
+  cursor: pointer;
+}
+
+.toggle-button svg {
+  transition: transform 0.3s ease;
+}
+
+@media (max-width: 991px) {
```
maxatdetroit: Whenever we have a media query for width, we should use SCSS instead and [use bootstrap media queries instead](https://getbootstrap.com/docs/5.3/layout/breakpoints/#media-queries) so that our breakpoints are consistent across the design system.

maxatdetroit (changes requested):
This component is already approx. 5 million times better than our experimental components. You're doing a fantastic job applying best practices. I added some comments here to take it to an even higher level of perfection.

src/stable/components/organisms/SectionNavigation/SectionNavigation.js
sreidthomas: I am going to run into the same issue with the other slots where I have to force styling with !

src/stable/components/organisms/SectionNavigation/SectionNavigation.js
sreidthomas: The underline is being controlled by bootstrap. I tried overriding it with this command but it did not work:
```
:host .section-nav ::slotted(a) {
  color: #000;
  text-decoration: none;
  border-bottom: none;
  padding: 0.25rem 0;
  display: block;
}

```
 How else could I get it to be text-decoration: none:
image

@maxatdetroit

sreidthomas: I also made the header a slot and now I cannot change the size of the header text. I'm not sure if bootstrap is overpowering my styles here as well: @maxatdetroit 

### MOBILE
image

### DESKTOP
image

src/stable/components/SectionNavigation/SectionNavigation.scss
```diff
@@ -0,0 +1,149 @@
...
+    }
+  }
+
+  :host .section-nav ::slotted(a) {
+    display: block;
+    padding: 0.75rem 1rem;
+    margin: 0;
+    border: none;
+    color: #000;
+    text-decoration: none;
```
maxatdetroit, line 141: As discussed offline, we should remove any typography styles from the slotted elements and let the lightDOM style the text. We can leave the layout styles though. This comment applies to the entire file, not just this example.

src/stable/components/SectionNavigation/SectionNavigation.js
```diff
@@ -0,0 +1,87 @@
...
+    const button = this.shadowRoot.querySelector('.toggle-button');
+    const container = this.shadowRoot.querySelector('.section-container');
+    if (button && container) {
+      const isExpanded = button.getAttribute('aria-expanded') === 'true';
+      button.setAttribute('aria-expanded', (!isExpanded).toString());
+      container.classList.toggle('expanded');
+    }
+  }
+
+  _wrapSlottedLinks() {
```
maxatdetroit, line 80: This function is not currently being called anywhere so the code is not being executed and the links assigned to the slots are not being wrapped. Would you call this in the constructor? It's good to set the slotchange event handler in the constructor in case the slot is assigned before the custom element is defined.

src/stable/stories/sectionnavigation.stories.js
```diff
@@ -0,0 +1,103 @@
...
+    await userEvent.click(toggleButton);
+    checkExpandedState(false);
+
+    // Verify navigation items
+    await waitFor(() => {
+      const slot = shadow.querySelector('slot[name="nav-items"]');
+      const navItems = slot.assignedElements();
+      expect(navItems.length).toBe(7);
+      expect(navItems[0].textContent).toBe('Services');
+      expect(navItems[0].getAttribute('href')).toBe('#services');
```
maxatdetroit: Would you also test the following scenarios:
1. slotted `a` elements are wrapped in `<li>`. This would have helped catch the bug I mentioned above with `_wrapSlottedLinks`.
2. non-`a` elements assigned to the slot are not used.

src/stable/components/SectionNavigation/SectionNavigation.js
```diff
@@ -0,0 +1,87 @@
...
+    // Wrap all assigned elements in <li> tags
+    const wrapLinks = () => {
+      const items = slot.assignedElements();
+      items.forEach((item) => {
+        // Only wrap <a> elements that are not already inside an <li>
+        if (item.tagName === 'A' && item.parentNode.tagName !== 'LI') {
+          const li = document.createElement('li');
+          item.parentNode.insertBefore(li, item);
+          li.appendChild(item);
+        }
```
maxatdetroit, line 107: We should also handle the case where someone passes something other than an `a` element. 

E.g. 

```
else if (item.tagName !== 'A') {
  item.remove();
}
```

maxatdetroit (changes requested):
You're like 99% there on how to control the slot elements. Left a couple comments to address the last 1%.

src/stable/stories/sectionnavigation.stories.js
```diff
@@ -0,0 +1,103 @@
...
+    const toggleButton = shadow.querySelector('.toggle-button');
+    await userEvent.click(toggleButton);
+    checkExpandedState(true);
+
+    // Test button click to collapse
+    await userEvent.click(toggleButton);
+    checkExpandedState(false);
+
+    // Verify navigation items
+    await waitFor(() => {
```
maxatdetroit, line 104: @SBBlee this test is a good example of how you would write your storybook component test for the tag element.

src/stable/components/SectionNavigation/SectionNavigation.scss
sreidthomas: Moved styles but cannot get some styles to work outside of the slots in the scss:

image

image

src/stable/components/SectionNavigation/SectionNavigation.js
```diff
@@ -74,12 +72,14 @@ class SectionNavigation extends HTMLElement {
...
           element.tagName === 'A' &&
-          !element.parentElement.classList.contains('nav-item')
+          !element.parentElement.tagName !== 'LI'
         ) {
           // If the condition is met, this creates a new <li> element
           const li = document.createElement('li');
           // adds the 'nav-item' class to the newly created <li> element
           li.classList.add('nav-item');
+          li.setAttribute('slot', 'nav-items');
+          element.removeAttribute('slot');
```
maxatdetroit, line 92: @sreidthomas here is the appropriate fix. Explanation below.

# Before the fix

The `setTimeout` was guaranteeing that the `_wrapSlottedLinks` was being called after all the slotchange events were triggered. Therefore, this `_wrapLinks` method was never actually being called (since everything was already in a slot). You can confirm this by looking at the DOM structure: notice how none of the links are in `li` elements. They're just the original `a` elements.

image

When we try to remove the `setTimeout`, the `_wrapLinks` gets called on the slot change but it removes the links from the ShadowDOM slot. Why? Because we're wrapping the links in a `li` element that does not have the `slot` attribute set, and ["only top-level children may have the slot="..." attribute"](https://javascript.info/slots-composition).

image

# After the fix

We remove the `setTimeout` call so that the `_wrapLinks` handler gets called on the initial slot change. We add the `slot="..."` attribute to the `li` item (which is the new top-level child) and remove the `slot="..."` attribute from the `a` item (which is no longer a top-level child).

src/stable/components/SectionNavigation/SectionNavigation.scss
maxatdetroit: I think we may be on a different page right now. 

I'm suggesting we move the typography styles (e.g. `font-size`) out of the `:host ::slotted(...)` selectors in the component SCSS (`SectionNavigation.scss`) and into the `<style>` element in the stories file (`sectionnavigation.stories.js`) instead. That will guarantee the slotted text elements are styled consistently with the rest of the website the component is used on.

I think you did the opposite in your recent commits.

sreidthomas: Yeah, I thought you wanted me to get rid of the slotted selectors completely @maxatdetroit

src/stable/components/SectionNavigation/SectionNavigation.js
```diff
@@ -0,0 +1,106 @@
+import styles from '!!raw-loader!./SectionNavigation.css';
+import varStyles from '!!raw-loader!../../../shared/variables.css';
+import bootstrapStyles from '!!raw-loader!../../../shared/themed-bootstrap.css';
+
+const template = document.createElement('template');
+template.innerHTML = `
+<style>
+${bootstrapStyles}
+${varStyles}
```
maxatdetroit: Do we need to include these styles? Does the section nav rely on these in any way or can we omit them?

src/stable/components/SectionNavigation/SectionNavigation.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  _handleToggle() {
+    const button = this.shadowRoot.querySelector('.toggle-button');
+    const container = this.shadowRoot.querySelector('.section-container');
+    if (button && container) {
+      const isExpanded = button.getAttribute('aria-expanded') === 'true';
+      button.setAttribute('aria-expanded', (!isExpanded).toString());
+      container.classList.toggle('expanded');
+    }
```
maxatdetroit: Similar to what you did for the gov banner:

This element should also have a public, boolean attribute 'expanded' that controls whether it is expanded or not.

src/stable/components/SectionNavigation/SectionNavigation.js
```diff
@@ -0,0 +1,106 @@
...
+        } else if (element.tagName === 'LI') {
+          // Handle li with anchor or li by itself
+          if (element.querySelector('a')) {
+            // li with anchor: ensure it has the correct class and slot
+            element.classList.add('nav-item');
+            element.setAttribute('slot', 'nav-items');
+          } else {
+            // li by itself: add a class but don't modify further
+            element.classList.add('nav-item');
+          }
```
maxatdetroit: We don't want to do this. This would mean we're allowing a nav item that is not a link.

src/stable/components/SectionNavigation/SectionNavigation.js
```diff
@@ -0,0 +1,106 @@
...
+    const wrapLinks = () => {
+      const assignedElements = slot.assignedElements();
+      assignedElements.forEach((element) => {
+        if (element.tagName === 'A' && element.parentElement.tagName !== 'LI') {
+          const li = document.createElement('li');
+          li.classList.add('nav-item');
+          li.setAttribute('slot', 'nav-items');
+          element.removeAttribute('slot');
+          element.parentNode.insertBefore(li, element);
+          li.appendChild(element);
```
maxatdetroit: With the exception of the if-statement, this is the same code as line 86. If this block gets executed, then the code at line 86 will never run (because its in an else). The `if` at line 69 is guaranteed to take precedence over line 86, because there's no case where the parent node of the slotted element is a `li`. The parent node is guaranteed to be the `ul` from the template because a slotted element is always a direct child.

src/stable/components/SectionNavigation/SectionNavigation.js
```diff
@@ -0,0 +1,106 @@
...
+        } else {
+          // Remove any other elements that are not li or a
+          element.remove();
+        }
```
maxatdetroit: Could we simplify this conditional logic a bit. It would be easier to read like this:

```
if (element.tagName === 'A') {
  // Wrap the 'a' element in an 'li'.
} else if (element.tagName === 'LI') {
  if (element.querySelector('a')) {
    // li with anchor: ensure it has the correct class and slot
  } else {
    // li without anchor: remove the element
  }
} else { 
  // Remove any other elements.
}
```

src/stable/stories/sectionnavigation.stories.js
```diff
@@ -0,0 +1,172 @@
...
+        >Standing Committees</a
+      >
+      <a slot="nav-items" href="#departments" class="section-nav-link"
+        >Departments, Contacts & More</a
+      >
+    </cod-section-navigation>
+  `,
+};
+
+export const SectionNavigationWithSpan = {
```
maxatdetroit: Since there is no test case on this story, we should remove it. We don't want people thinking they should pass spans in the slot.

maxatdetroit (changes requested):
Some final tweaks suggested.

maxatdetroit (approved):
its-happening

PR #289, opened 2025-02-11
Feature.Tag

SBBlee:
Adds Tag Component to Design System

image

https://github.com/user-attachments/assets/b9058dfd-4651-47ed-8552-59e9b62772f3


maxatdetroit:
@sreidthomas would you take the initial review pass on this PR? Both of you are officially design system experts. See what you notice should be tweaked / improved. I can take a review pass after you.

src/experimental/components/atoms/Tag/Tag.js
sreidthomas, line 1: For the new design system set, we are using a stable folder. We're also no longer using the atoms, molecules, organisms naming system. You can just create a folder with the name of the component, in your case, Tag, and add it to the components folder. Here is my PR. You can see the folder structure set up on the right: https://github.com/CityOfDetroit/COD-Design-System/pull/282/files

sreidthomas: When you change the folder structure, you will also have to update the imports because the project is going to be looking for those files in the wrong place.

src/experimental/stories/tag.stories.js
```diff
@@ -0,0 +1,49 @@
+import '../components/atoms/Tag/cod-tag';
+
+// Function to strip HTML tags
+function stripHTML(html) {
+  const doc = new DOMParser().parseFromString(html, 'text/html');
+  return doc.body.textContent || "";
+}
+
+export default {
+  title: 'Experimental/Atoms/Tag',
```
sreidthomas: When you add the Stable folder and update the directory, you'll have to change this line as well: Stable/Tag

src/experimental/components/atoms/Tag/Tag.js
```diff
@@ -0,0 +1,51 @@
+// Define a template element to hold the structure and style of the custom element
+const template = document.createElement('template');
+template.innerHTML = `
+  <div class="tag-container d-inline">
```
sreidthomas: You're using inline, bootstrap styles. I do believe we are supposed to use CSS / SCSS for our stylings.

src/experimental/stories/tag.stories.js
```diff
@@ -0,0 +1,49 @@
+import '../components/atoms/Tag/cod-tag';
+
+// Function to strip HTML tags
```
sreidthomas: This function isn't needed. You were already handling the slotchange event and stripping HTML tags in your Tag.js file. I would refer back to Max's example for better understanding: https://gist.github.com/maxatdetroit/f8fd9a0f78c732a043812d4ce9189601#file-example-controlled-slots-html-L51-L57

src/experimental/stories/tag.stories.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export default {
+  title: 'Experimental/Atoms/Tag',
+  argTypes: {
+    label: { control: 'text', defaultValue: 'Label' },
+  },
+};
+
+// Template
```
sreidthomas: Lines 16-31 are repeating the code you already have in your Tag.js file. These would be the implementation details of the component, which should stay in the Tag.js file. The stories file should focus on demonstrating how to use the components. So these lines are not needed here.

src/experimental/components/atoms/Tag/Tag.js
sreidthomas, line 1: Also, to fix the prettier build fail, you need to run: yarn lint. It will show you the files that have the issues and then you run: yarn prettier -w "COPY THE FILE THAT FAILED HERE"

SBBlee:
@sreidthomas Thanks for the help on my PR! I'm working on it.

src/stable/stories/tag.stories.js
```diff
@@ -0,0 +1,28 @@
+import '../components/atoms/Tag/cod-tag';
+
+export default {
+  title: 'Stable/components/Atoms/Tag',
```
maxatdetroit: We're not using the atoms/molecules/organisms folder structure anymore. We can just put `Stable/Tag` here.

src/stable/components/atoms/Tag/tag.scss
maxatdetroit, line 1: Remove the `atoms` directory from the path please. It should just be `src/stable/components/Tag/tag.scss`

src/experimental/stories/tag.stories.js
maxatdetroit, line 1: You can delete this file now. `git rm src/experimental/stories/tag.stories.js`

src/experimental/components/atoms/Tag/cod-tag.js
maxatdetroit, line 1: ditto re: delete this file

src/experimental/components/atoms/Tag/Tag.js
maxatdetroit, line 1: ditto re: delete this file

src/stable/components/atoms/Tag/Tag.js
maxatdetroit, line 1: ditto: re: remove `atoms` from the path.

src/stable/components/atoms/Tag/Tag.js
```diff
@@ -0,0 +1,58 @@
...
+  _setupSlot() {
+    // Find the slot element and add an event listener for the slotchange event
+    const slot = this.shadowRoot.querySelector('slot[name="label"]');
+    slot.addEventListener('slotchange', () => {
+      const elements = slot.assignedElements(); // Get the elements assigned to the slot
+      elements.forEach(element => {
+        // Check if the element is not a <span> or <a>
+        if (element.tagName !== 'SPAN' && element.tagName !== 'A') {
+          // Strip inner HTML content
+          element.innerHTML = stripHTML(element.innerHTML);
```
maxatdetroit: What is this line of code **supposed** to do? I don't think it's doing quite what you want. 

I'm assuming this slotchange event handler is supposed to:
1. catch the case when someone assigns anything other than an `a` or `span` element to the `label` slot
2. force the assigned elements to use `a` or `span` slots instead

Currently (1) is happening, but (2) is not. Consider the following cases and how this code would behave:
1. Someone assigns a `<h1>` element to the slot. What should happen: the assigned element becomes a `span` element instead of `h1` and the innerText of the assigned element is unchanged. That's not currently happening.
2. Someone assigns a `<ul>` element with children list items. What should happen: the assigned element becomes a `span` element instead of `ul` with nested children and the innerText of the assigned element is set to the innerText of the list. That's not currently happening. 

You can use the browser javascript console to test the event handler code given different assigned elements. It's a little easier than trying to debug this code while it runs in storybook:

```
function stripHTML(html) {
  const doc = new DOMParser().parseFromString(html, 'text/html');
  return doc.body.textContent || "";
}

const element1 = document.createElement('a')
element1.innerText = 'Example Link'

console.log(element1.outerHTML)
<a>Example Link</a>
stripHTML(element1.innerHTML)
"Example Link"

const element2 = document.createElement('ul')
const listItem = document.createElement('li')
listItem.innerText = 'Example List Item'
element2.appendChild(listItem)

console.log(element2.outerHTML)
<ul><li>Example List Item</li></ul>
console.log(element2.innerHTML)
<li>Example List Item</li>
element2.innerHTML = stripHTML(element2.innerHTML)
"Example List Item"

console.log(element2.outerHTML)
<ul>Example List Item</ul>
```

src/stable/components/atoms/Tag/Tag.js
```diff
@@ -0,0 +1,58 @@
+import styles from '!!raw-loader!./tag.css';
+
+// Function to strip HTML tags
+function stripHTML(html) {
+  const doc = new DOMParser().parseFromString(html, 'text/html');
+  return doc.body.textContent || "";
+}
```
maxatdetroit: Using `innerText` (https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText) would be more appropriate for what we're trying to accomplish. See my comment below for more details.

maxatdetroit (changes requested):
This is really, really close. We just have to fix the slotchange event handler to properly handle the cases when someone assigns elements other than `a` and `span`. 

Also, we need to write some [Storybook tests](https://storybook.js.org/docs/7/writing-tests/interaction-testing) to test our expectations about how the component is behaving in different scenarios. Specific scenarios to test:
1. Someone assigns a `a` element to the label slot. Expectation to check: an `a` element is present in the label slot.
2. Someone assigns a `ul` element with children `li` elments to the label slot. Expectation to check: a `span` element is present in the label slot with the span innerText made up of the list innerText.

src/stable/components/atoms/Tag/Tag.js
maxatdetroit: As discussed over Teams, this is how the slotchange handler should look:

```js
slot.addEventListener('slotchange', () => {
  const elements = slot.assignedElements(); // Get the elements assigned to the slot
  elements.forEach(element => {
    // Check if the element is not a <span> or <a>
    if (element.tagName !== 'SPAN' && element.tagName !== 'A') {
      // OLD CODE:
      // Strip inner HTML content
      // element.innerHTML = stripHTML(element.innerHTML);
      // NEW, CORRECT CODE:
      const spanElement = document.createElement('span');
      spanElement.innerText = element.innerText;
      element.replaceWith(spanElement);
    }
  });
});
```

You can see how this would handle the different scenarios I mentioned above.

src/stable/components/Tag/Tag.js
```diff
@@ -0,0 +1,80 @@
...
+    slot.addEventListener('slotchange', () => {
+      // Get the elements assigned to the slot
+      const elements = slot.assignedElements();
+      // Define allowed tags
+      const allowedTags = ['SPAN', 'A'];
+      
+      // Iterate over the assigned elements
+      elements.forEach((element, index) => {
+        // Check if the element is an <li>
+        if (element.tagName === 'LI') {
```
maxatdetroit: There's not really a need to check for if the element is an `li` element. 

Let's think about the possible elements and how this function should behave:

1. The element is anything other than `a` or `span` -> convert it to a `span` (already happening at line 63-69)
2. The element is a `a` or `span` -> do nothing (we don't need code for this case, its implied)

After all that, we know the element is either an `a` or a `span`, and we can just append a comma between each element. So this code doesn't need to check what type of element it is.

src/stable/components/Tag/Tag.js
```diff
@@ -0,0 +1,80 @@
...
+          // Create a span element to replace the <li>
+          const spanElement = document.createElement('span');
+          spanElement.slot = 'label';
+          spanElement.innerText = element.innerText.replace(/<br>/g, '').replace(/\n/g, '');  // Removing <br> and newlines
+          
+          // Append a comma between every <li> 
+          if (index < elements.length - 1) {
+            spanElement.innerText += ',';
+          }
```
maxatdetroit: I'm basically saying you should move this logic to the end of the `forEach` loop and remove this `if (element.tagName === 'LI') {` block entirely.

src/experimental/stories/event.stories.js
maxatdetroit, line 1: This needs to be removed from this PR, otherwise the GitHub checks are going to continue to fail.

maxatdetroit (changes requested):

src/stable/components/Tag/Tag.js
```diff
@@ -0,0 +1,69 @@
+import styles from '!!raw-loader!./tag.css';
```
maxatdetroit: This filename is typically capitalized (just a standard we've seem to adopted in the design system).

src/stable/components/Tag/Tag.js
```diff
@@ -0,0 +1,69 @@
...
+          }
+        }
+      });
+    });
+  }
+
+  isTagConnected() {
+    // Check if the tag-container is connected to the shadow DOM
+    return this.shadowRoot.querySelector('.tag-container') !== null;
+  }
```
maxatdetroit: This isn't used anywhere, and I don't think it's needed.

src/stable/components/Tag/tag.scss
```diff
@@ -0,0 +1,3 @@
+.tag-container {
```
maxatdetroit: Given how simple this CSS is, there's really no need to have a `scss` file for this stylesheet. You could delete the `.css.map` file and the `.scss` file and leave just the `.css` file. That would help simplify the codebase.

maxatdetroit (changes requested):
This is spot on. I left a couple minor comments to address inline. Also, would you look over best practice (6) in #201 and update this PR to have similar documentation?

maxatdetroit (approved):
giphy

PR #290, opened 2025-02-12
ServiceButton Component Creation

### LINKED ISSUE #283: Create service button component
# Service Button Component

The goal of this task is to create a new service button component in the stable build of the design system.

## Mockups

Image

Note: Default state on the left, hover state on the right.

### Mobile

Image

### Desktop

Image

## Requirements

- Minimum width of the component should be 340px.
- Transition between hover states should be a 200ms ease-in-out transition.
- The spirit SVG in the hovered state is here: Detroit Spirit
- The title and subtitle of the button should be slotted elements, however they need to be tightly controlled. We don't want the user of the service button to dictate the parent element of the title and subtitle. To accomplish this, we should use the [slotchange event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/slotchange_event) to strip away any HTML tags the user may have placed in the slot and just take the text content. E.g.: https://gist.github.com/maxatdetroit/f8fd9a0f78c732a043812d4ce9189601

sreidthomas:
### Mobile

Image

### Desktop

Image

## Requirements

- [ ] Minimum width of the component should be 340px.
- [ ] Transition between hover states should be a 200ms ease-in-out transition.
- [ ] The spirit SVG in the hovered state is here: Detroit Spirit
- [ ]  The title and subtitle of the button should be slotted elements, however they need to be tightly controlled.
- [ ]  use the [slotchange event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/slotchange_event) to strip away any HTML tags the user may have placed in the slot and just take the text content.

----------------------------------------------------------------

### **BUTTON WITHOUT HOVER:**
image
image
image

### **BUTTON WITH HOVER:**

image
image
image

### **ADD TESTS FOR COMPONENT**

- [x] Test for title and subtitle:
    Verify that the title text is rendered correctly
    Verify that the subtitle text is rendered correctly
- [x] Test for hover:
    Verify that the button changes color on hover
    Check that the overlay effect appears on hover
- [ ]Test for hover with SVG:
    Verify that the SVG appears on hover
    Check that the SVG is positioned correctly (bottom right)

### **TEST RESULTS:**
image

image


src/stable/components/ServiceButton/ServiceButton.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+
+    subtitleSlot.addEventListener('slotchange', () => {
+      const elements = subtitleSlot.assignedElements();
+      if (elements.length > 0) {
+        const text = elements[0].textContent;
+        elements[0].textContent = text;
+      }
+    });
+  }
```
maxatdetroit: See my comment on Shannon's PR for how these slotchange handlers should work to restrict the elements allowed. Basically, if the assigned element is anything other than a `span` we should just replace it with a span and copy over the originally assigned innerText.

src/stable/components/ServiceButton/ServiceButton.js
```diff
@@ -0,0 +1,53 @@
+import styles from '!!raw-loader!./ServiceButton.css';
+import varStyles from '!!raw-loader!../../../shared/variables.css';
+import bootstrapStyles from '!!raw-loader!../../../shared/themed-bootstrap.css';
+
+const template = document.createElement('template');
+
+template.innerHTML = `
+<style>
+${bootstrapStyles}
+${varStyles}
```
maxatdetroit: Try getting rid of these like we did in the other PR. We should avoid adding unnecessary style sheets whenever possible.

src/stable/components/ServiceButton/ServiceButton.scss
```diff
@@ -0,0 +1,3 @@
+@import 'bootstrap/scss/functions';
+@import 'bootstrap/scss/variables';
+@import 'bootstrap/scss/mixins';
```
maxatdetroit: giphy

Is this missing the rest of the styles you created?

src/stable/stories/servicebutton.stories.js
```diff
@@ -0,0 +1,104 @@
...
+
+    // Test for title and subtitle
+    const title = shadow.querySelector('.title slot');
+    const subtitle = shadow.querySelector('.subtitle slot');
+    await expect(title.assignedNodes()[0].textContent).toBe('Apply for a Job');
+    await expect(subtitle.assignedNodes()[0].textContent).toBe(
+      'View job postings for the City of Detroit or our partners.',
+    );
+
+    // Test for hover effect
```
maxatdetroit: This isn't actually testing the hover... can you write something to check that the hover styles are applied?

src/stable/stories/servicebutton.stories.js
```diff
@@ -0,0 +1,104 @@
...
+    <cod-service-button>
+      <span slot="title">Apply for a Job</span>
+      <span slot="subtitle"
+        >View job postings for the City of Detroit or our partners.</span
+      >
+    </cod-service-button>
+  `,
+  play: async ({ canvasElement }) => {
+    // Allow initial render
+    await new Promise((resolve) => setTimeout(resolve, 100));
```
maxatdetroit: Is this necessary? We haven't needed to do this on other component tests.

src/stable/stories/servicebutton.stories.js
```diff
@@ -0,0 +1,104 @@
...
+
+    const serviceButton = canvasElement.querySelector('cod-service-button');
+    const shadow = serviceButton.shadowRoot;
+
+    // Test for title and subtitle
```
maxatdetroit: Could we also add a test that tries to assign non-span elements and check that they get converted to spans?

src/stable/stories/servicebutton.stories.js
```diff
@@ -0,0 +1,104 @@
...
+    // Test for style changes
+    const computedStyle = window.getComputedStyle(button);
+    const hoverState = button.matches(':hover');
+
+    // Check if any of these conditions are met
+    const styleChangeDetected =
+      styleChanged || // MutationObserver detected change
+      hoverState || // Element is in hover state
+      computedStyle.getPropertyValue('--hover-opacity') !== '' || // Custom property changed
+      parseFloat(computedStyle.getPropertyValue('opacity')) > 0; // Opacity changed
```
maxatdetroit: Shouldn't these be `&&` instead of `||`? All of these things should be true.

src/stable/stories/servicebutton.stories.js
```diff
@@ -0,0 +1,104 @@
...
+
+    expect(styleChangeDetected).toBe(true);
+
+    // Cleanup
+    observer.disconnect();
+
+    // Test icon if present
```
maxatdetroit: typo. 'if' should be 'is'.

src/stable/components/ServiceButton/ServiceButton.js
```diff
@@ -0,0 +1,53 @@
...
+${styles}
+
+</style>
+<button>
```
maxatdetroit: I forgot to mention this in the issue: the service button needs to have an `href` attribute and the button should be a link element (not a button element). We should also add a test for this (to check that the button is a link and can be clicked).

maxatdetroit (changes requested):
This is definitely on the right track. We're going to have to think about how to handle accessibility requirements (e.g. aria attributes) inside the component, so feel free to think about that yourself as you're revising this PR. I'll also review for accessibility when I come back on the next PR review.

src/stable/components/ServiceButton/ServiceButton.js
sreidthomas: when I remove these my styles no longer work as they did. The svg does not show. So how are these not being used @maxatdetroit?

src/stable/components/ServiceButton/ServiceButton.scss
sreidthomas: I never added to scss. I was testing in css to not have to compile everytime I made a change. that might have been a mistake though @maxatdetroit

src/stable/stories/servicebutton.stories.js
maxatdetroit: Sorry, I meant to delete this comment before submitting. I see the hover tested below. Ignore this comment please.

src/stable/components/ServiceButton/ServiceButton.js
maxatdetroit: This comment is just referencing lines 9-10 which are the `bootstrapStyles` and `varStyles`. The `styles` variable is needed, but I wouldn't expect the `bootstrapStyles` and `varStyles` to be necessary. 

> So how are these not being used?

I would ask you to prove whether they are being used or not. We have to dig deeper than whether the SVG shows or not. Why doesn't the SVG show? If we inspect the component HTML/CSS in the browser, do we see the SVG is dependent on styles in `bootstrapStyles` or `varStyles`? If not, we can remove those specific styles and solve the issue elsewhere.

sreidthomas:
### **STILL TO DO:**

- [ ]  We should also add a test for this (to check that the button is a link and can be clicked).

src/stable/components/ServiceButton/ServiceButton.scss
maxatdetroit: It's fine to leave everything in CSS for this component and not use SCSS at all. The stylesheets are simple enough to where they wouldn't benefit from SCSS. 

I was just confused why this file exists. We should delete it if we're not using it, otherwise the next time someone runs the SCSS compiler, it's going to overwrite your CSS. Also, this empty file will confuse simple-minded people like me when they read the code. ;)

sreidthomas:
I kept having issues with the other 2 tests I had. Do you have any test ideas? I was trying mock click, which kept clicking the button and sending it to the dummy link whenever I opened Storybook. And the other was check for something other than span it kept changing the layout of the component: @maxatdetroit 
test4results

src/stable/stories/servicebutton.stories.js
```diff
@@ -0,0 +1,63 @@
...
+
+export default {
+  title: 'Stable/ServiceButton',
+  component: 'cod-service-button',
+  tags: ['autodocs'],
+};
+
+export const ServiceButton = {
+  render: () => html`
+    <cod-service-button>
```
maxatdetroit: We need to support the attributes that an `a` element supports. Specifically: `href`, `hreflang`, `rel`, and `target`. These will control the attributes of the `a` element in your component template.

Right now, the button always links to `example.com` which is not what we want.

src/stable/components/ServiceButton/ServiceButton.scss
```diff
@@ -0,0 +1,62 @@
...
+
+a.service-button {
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: flex-start;
+  padding: 50px 25px;
+  background-color: #004445;
+  border-radius: 8px;
```
maxatdetroit: As discussed: these should be square corners, not rounded. See the Figma mockups for exact details of the designs.

src/stable/stories/servicebutton.stories.js
```diff
@@ -0,0 +1,63 @@
...
+    expect(subtitleNodes[0].textContent).toBe(
+      'View job postings for the City of Detroit or our partners.',
+    );
+
+    // ===== TEST 2: Link Element Test =====
+    // Verifies that the component contains a link element with href attribute
+    const link = shadow.querySelector('a');
+    expect(link).not.toBeNull();
+    expect(link.tagName).toBe('A');
+    expect(link.hasAttribute('href')).toBe(true);
```
maxatdetroit, line 70: We should update this test to check the link attributes are identical to those set on the component (once we add support for those). That would have caught the issue I commented on at line 13.

maxatdetroit (changes requested):
This is pretty close to done but needs a few tweaks.

maxatdetroit (approved):
its-happening

PR #310, opened 2025-03-14
Icon Component Update

### LINKED ISSUE #295: Update icon component
# Icon Component

The goal of this task is to update the existing a icon component in the experimental build of the design system and move it into the stable build.

The update icon should have the following changes:
- We should only use icons from Bootstrap https://icons.getbootstrap.com/
- The only attributes on the icon should be
  - `name` (of the icon to use)
  - `label` (for adding an aria-label to the icon used by a11y tech)
- The color and size of the icon should be determined by the `color` and `font-size` style properties (either applied directly to the icon element or its parent). See an example here: https://shoelace.style/components/icon#colors

Note: We need to support "backwards compatibility" for cases where someone is already using the icon element in production. To do so, the old attributes (e.g. `data-size`) must remain usable and, when provided, behave the same as previously.

sreidthomas:
This PR is for updating the Icon Component that exists in the experimental by:

- [x] Moving Icon from experimental to stable folder
- [x] Adding bootstrap icons and the backwards compatibility use of font awesome (some icons do not have bootstrap icons)
- [x] Adding new name / label attributes
- [x] Keeping the old data-icon, data-size and is-highlighted attributes for backwards compatibility
- [x] The color and size of the icon should be determined by the color and font-size style properties (either applied directly to the icon element or its parent). See an example here: https://shoelace.style/components/icon#colors


maxatdetroit:
@sreidthomas I removed myself from review for now since I think you're still working on this. When you're ready for review, go ahead and re-add me as a reviewer.

sreidthomas:
It's saying conflicts for src/experimental/stories/icon.stories.js but I already moved that file. It's no longer in that location @maxatdetroit

src/experimental/index-experimental.js
```diff
@@ -23,7 +23,7 @@ import './components/atoms/FormCheck/cod-formcheck';
 import './components/atoms/FormControl/cod-formcontrol';
 import './components/atoms/FormLabel/cod-formlabel';
 import './components/atoms/FormSelect/cod-formselect';
-import './components/atoms/Icon/cod-icon';
+import '../stable/components/Icon/cod-icon';
```
maxatdetroit: This import needs to be moved from the `src/experimental/index-experimental.js` to `src/stable/index-stable.js`.

src/stable/components/Icon/Icon.js
```diff
@@ -0,0 +1,622 @@
+import styles from '!!raw-loader!./Icon.css';
```
maxatdetroit, line 1: For future reference when moving and refactor files in a branch, if you follow this order of operations, it makes the PR easier to review. Instead of recording the change as a file deletion with an entirely new file, git will record the change as moving a file then changing the moved file:

1. `git mv <old-file-path> <new-file-path>`
2. `git commit -m "Moved file path"
3. <make all the edits here to the file in the new path/location>
4. `git commit -m "Refactor file"`

It's a little hard to see how this component changed now because git thinks its an entirely new file so there's no side-by-side diff.

src/stable/components/Icon/Icon.js
```diff
@@ -0,0 +1,622 @@
...
+      return;
+    }
+    this.renderIcon();
+  }
+
+  isIconConnected() {
+    return this.shadowRoot.querySelector('.icon').innerHTML !== '';
+  }
+
+  renderIcon() {
```
maxatdetroit: Let's make this a 'private method' like `_renderIcon()`.

src/stable/components/Icon/Icon.js
```diff
@@ -0,0 +1,622 @@
...
+  }
+
+  connectedCallback() {
+    if (this.isIconConnected()) {
+      return;
+    }
+    this.renderIcon();
+  }
+
+  isIconConnected() {
```
maxatdetroit, line 61: ditto re: making it a private method

src/stable/components/Icon/Icon.js
```diff
@@ -0,0 +1,622 @@
...
+    // Set icon using getIcon()
+    iconElement.innerHTML = this.getIcon(icon, size, library);
+
+    // Handle boolean attribute
+    if (this.hasAttribute('is-highlighted')) {
+      container.classList.add('highlighted');
+    }
+  }
+
+  getIcon(icon, size, library = 'fontawesome') {
```
maxatdetroit: ditto re: making this a private method

maxatdetroit: Also, could we move this method into a separate helper function. E.g. making a new file `icon-helpers.js` in the `src/stable/components/Icon/` folder that just has this method exported. Then import the method for use in this component. This method is going to get crazy long so it'll be easier to maintain the SVGs in another file, separate from the core logic of this custom element.

src/stable/components/Icon/Icon.js
```diff
@@ -0,0 +1,622 @@
...
+  }
+
+  renderIcon() {
+    const container = this.shadowRoot.querySelector('.icon-container');
+    const iconElement = this.shadowRoot.querySelector('.icon');
+
+    // Get attributes
+    const icon = this.getAttribute('name') || this.getAttribute('data-icon');
+    const label =
+      this.getAttribute('label') || this.getAttribute('aria-label') || icon;
```
maxatdetroit: We should just use `label`, not `aria-label` as a fallback. `aria-label` has a very specific meaning and we don't want it applied to the custom element itself (only the icon container like you do below).

src/stable/components/Icon/Icon.js
```diff
@@ -0,0 +1,622 @@
...
+    let size =
+      this.getAttribute('size') || this.getAttribute('data-size') || '24';
```
maxatdetroit: We don't want to introduce a new `size` attribute. As mentioned in #295, the new way for users to set the size is to set the `font-size` in the lightDOM. But the `data-size` attribute should also continue to work as previously.

maxatdetroit (changes requested):
In addition to the inline comments below, how are you testing this? If you're testing manually using storybook, then would you please list the scenarios you tested on the PR and include a video of the testing?

Note: There are a lot of scenarios to test given the scope of changes here. Make sure to test all the new combinations. E.g. 
- `data-size` and lightDOM `font-size` are set; what happens? only one is set; what happens? etc.
- `data-icon` and `name` attribute are set; what happens? only one is set; what happens? etc.
- ... same thing with `library` attribute.

Also, please see best practices (4) and (6) in #201 . We should update the Icon to reflect JS properties from the `name` attribute and improve our storybook documentation.

src/stable/components/Icon/Icon.js
sreidthomas: I am not sure I understand @maxatdetroit

src/stable/components/Icon/Icon.js
```diff
@@ -18,6 +18,56 @@ class Icon extends HTMLElement {
     this.shadowRoot.appendChild(template.content.cloneNode(true));
   }
 
+  // Define reflective properties
+  static get observedAttributes() {
+    return ['name', 'library', 'data-size', 'label', 'is-highlighted'];
+  }
+
```
maxatdetroit: See my explanation on #318 regarding reflective properties and a clean way to implement both reflective and non-reflective properties. 

In this component, the only reflect properties should be `size`, `name`, and `library`.

maxatdetroit (changes requested):
The property implementation needs to be tweaked slightly, otherwise this is on the right track.

src/stable/components/Icon/Icon.js
sreidthomas: I am not sure how to do it for this one. When I remove label and is-highlighted the component does not function as it should @maxatdetroit

jedgar1mx:
Need story to review inheritance on both size and color

src/stable/components/Icon/Icon.js
```diff
@@ -0,0 +1,137 @@
...
+    return this.hasAttribute('is-highlighted');
+  }
+
+  set isHighlighted(value) {
+    if (value) {
+      this.setAttribute('is-highlighted', '');
+    } else {
+      this.removeAttribute('is-highlighted');
+    }
+  }
```
maxatdetroit: `is-highlighted` shouldn't be reflected, so it shouldn't have a JS property setter method.

src/stable/components/Icon/Icon.scss
```diff
@@ -0,0 +1,28 @@
+:host {
+  // Default to inheriting color from parent
+  --icon-color: inherit;
+  display: inline-block;
+}
+
+// /* Highlighted styles */
+.icon-container {
+  display: inline-block;
+  color: var(--icon-color);
```
maxatdetroit: Could we just put `color: inherit` here and remove the custom property?

src/stable/docs/Icon.mdx
```diff
@@ -0,0 +1,85 @@
...
+  The `Icon` component displays an icon from a specified icon library.
+</div>
+
+## Examples
+
+### Template
+
+<Description of={IconStories.IconArguments} />
+<Story of={IconStories.IconArguments} inline={true} />
+<Source of={IconStories.IconArguments} />
```
maxatdetroit: Could we include the controls in the documentation?

src/stable/docs/Icon.mdx
```diff
@@ -0,0 +1,85 @@
...
+{ "name": "icon", "description": "Optional slot to manually provide a custom icon or SVG if not using the built-in icon system."}
+])}>
+</doc-slots-table>
+
+## HTML Attributes / JS Properties
+
+Supported HTML attributes. If the attribute is reflected in a JS property, that attribute will have a 'reflects' tag next to the name.
+<doc-js-properties-table
+data={JSON.stringify([
+{ "name": "data-icon", "description": "Sets the icon name to display. Often used for dynamic rendering or legacy compatibility.", "type": "string", "default": "-","reflects": false},
```
maxatdetroit: In the documentation, could we group the duplicated attributes. E.g. put `data-icon` right after `name` in the table.

Also, let's clearly mark all the old attributes as DEPRECATED by putting `*DEPRECATED*` as the first word in the description of those fields.

maxatdetroit: Also, some of these reflect but they're marked as non-reflective in these docs.

src/stable/docs/Icon.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+# Icon
+
+<div className="component-summary">
+  The `Icon` component displays an icon from a specified icon library.
+</div>
+
+## Examples
+
+### Template
```
maxatdetroit, line 26: We should add a section in the docs to describe how to set size and color using the parents CSS.

maxatdetroit (changes requested):
When I try to view the inheritance stories for testing I'm receiving `undefined` errors in Storybook. Is this just on my end, or are you seeing the same issues @sreidthomas? 

image

I also left some comments to fix before we merge this in.

sreidthomas:
> When I try to view the inheritance stories for testing I'm receiving `undefined` errors in Storybook. Is this just on my end, or are you seeing the same issues @sreidthomas?
> 
> image
> 
> I also left some comments to fix before we merge this in.
I am getting the same. I added this update as per Edgar's review and I have not been sure how to get it to work without the icons being undefined.

sreidthomas:
> > When I try to view the inheritance stories for testing I'm receiving `undefined` errors in Storybook. Is this just on my end, or are you seeing the same issues @sreidthomas?
> > image
> > I also left some comments to fix before we merge this in.
> > I am getting the same. I added this update as per Edgar's review and I have not been sure how to get it to work without the icons being undefined.
### **THIS IS FIXED**

src/stable/components/Icon/Icon.scss
```diff
@@ -0,0 +1,28 @@
+:host {
+  // Default to inheriting color from parent
+  --icon-color: inherit;
```
maxatdetroit, line 3: This property doesn't get used anywhere does it? Can we remove it?

maxatdetroit: This property doesn't get used anywhere does it? Can we remove it here and where it gets set in `Icon.js`?

maxatdetroit (changes requested):
It looks like the color is being inherited but not the size. Notice in the video that 48px size and 64px have the same sized icon. Do you know what the issue is? Would you please see if you can fix it? Let me know if you want a hand.


https://github.com/user-attachments/assets/2fbf1cd6-da86-4665-bba2-26803df1611c

sreidthomas:
> It looks like the color is being inherited but not the size. Notice in the video that 48px size and 64px have the same sized icon. Do you know what the issue is? Would you please see if you can fix it? Let me know if you want a hand.
> Screen.Recording.2025-05-06.102319.mp4

Yeah I'm not sure if you remember but we spoke about this and I couldn't figure out why the size was not being inherited only the color. I did ask for help with this @maxatdetroit

PR #318, opened 2025-03-25
ProfileCard component

### LINKED ISSUE #294: Create profile component
# Profile Component

The goal of this task is to create a new profile component in the stable build of the design system.

## Mockups

The design should look identical on desktop and mobile:

Image

The following examples are to demonstrate how the component should behave in its parent containers in different viewport scenarios.

### Mobile

Image

### Desktop

Image

## Requirements

- The image src URL should be provided as an attribute on the component.
- The name of the individual should be a named slot. The assigned element should be either a link or a span element.
- The title(s) of the individual (e.g. Sandra Yu Stahl is both "Deputy Chief Financial Officer" and "Chief Procurement Officer") should be provided as additional named slotted elements. The assigned element should be a span.

sreidthomas:
The goal of this task is to create a new profile component in the stable build of the design system.

ProfileCard should be behave the same on Desktop & Mobile:

### Mobile

Image

### Desktop

Image

## Requirements

- [x] The width of the component (and component contents) should scale with the parent container

- [x] The image src URL should be provided as an attribute on the component.

- [x] The name of the individual should be a named slot. The assigned element should be either a link or a span element.

- [x] The title(s) of the individual (e.g. Sandra Yu Stahl is both "Deputy Chief Financial Officer" and "Chief Procurement Officer") should be provided as additional named slotted elements. The assigned element should be a span.

- [x] Follow the best practices for attributes and slots described in #201 , #202 , and #205 

### **FINAL RESULT:**

https://github.com/user-attachments/assets/a8fbd0f0-7637-49cd-b1c4-b3ec82ece24e

image


src/experimental/cod-gov-banner.js
maxatdetroit, line 1: giphy

This shouldn't be here.

src/stable/components/ProfileCard/ProfileCard.js
```diff
@@ -0,0 +1,222 @@
...
+          firstElement instanceof HTMLSpanElement ||
+          firstElement instanceof HTMLAnchorElement
+        )
+      ) {
+        throw new Error(
+          'ProfileCard: The "name" slot should contain either a <span> or an <a> element.',
+        );
+      }
+    }
+  }
```
maxatdetroit: I was talking to @SBBlee about this: I think we should actually stop adding in these "sanity checks" to force people to use the components the right way. We should just update the documentation to tell people how to use the component, and if they misuse it, that's on them.

So you can go ahead and remove these 'validate slot' assertion methods and their invocations.

src/stable/components/ProfileCard/ProfileCard.js
```diff
@@ -0,0 +1,222 @@
...
+
+  connectedCallback() {
+    this._updateImage(this.src);
+    this._updateLink(this.href);
+    this._validateNameSlot();
+  }
+
+  getValidationError() {
+    return this._validationError;
+  }
```
maxatdetroit: I don't think we use this.

src/stable/components/ProfileCard/ProfileCard.js
```diff
@@ -0,0 +1,222 @@
...
+    this.setAttribute('src', value);
+  }
+
+  get href() {
+    return this.getAttribute('href');
+  }
+
+  set href(value) {
+    this.setAttribute('href', value);
+  }
```
maxatdetroit: "Fully" reflected would mean changes to the attribute automatically update the property and vice versa. Right now, these properties are "partially" reflected: changes to the property automatically update the attribute, but not vice versa. That would cause some unexpected behavior, like this:
```js
const card = createElement('cod-profile-card').setAttribute('href', 'foo');
console.log(card.href); // output: foo
card.href = 'bar';
console.log(card.href); // output: bar
card.setAttribute('href', 'baz');
console.log(card.href); // output: bar (not baz)
```

We want these properties to be completely non-reflective. Completely non-reflective would mean changes to the attribute automatically reflect on the JS property, but not vice versa. How would you go about that cleanly? 
- Remove the setter method for the properties
- Create an internal property for managing the state of each variable (e.g. `this._href`)
- Always use the internal properties when rendering the element / updating the element DOM
- When the attribute is changed, update the internal property, then re-render the component DOM

You can see an example of a non-reflected JS property `href` here on the button component: https://github.com/CityOfDetroit/COD-Design-System/blob/bv3.0/src/stable/components/Button/Button.js

Note: I forget to add a getter method for href on that button, so ignore the fact that the getter is missing.

maxatdetroit (changes requested):
This is definitely moving in the right direction. Some minor things:
- When the profile card is a link (i.e. when the `href` attribute is provided), would you make sure that a right caret / chevron is added at the end of the person's name? See the mobile mockup screenshot for an example of what I'm talking about.
- The vertical margins / padding between the text elements is off from the mockups. Would you go into figma and see what the right design specs are for that and mimic them here?
- Would you update the inline styles in the stories to remove the blue text color on the profile card links in storybook?
- Would you update the documentation page to include a section at the top for 'Usage' that includes Storybook controls? See an example of what I mean here: https://cityofdetroit.github.io/COD-Design-System/?path=/docs/components-tag--tag#usage

src/experimental/cod-gov-banner.js
sreidthomas: Oh yeah that's what I been meaning to show / ask / tell you. For some reason, when I run yarn lint, it keeps adding files from my other components. I've been removing them but by the time I realized it was doing it I probably had already pushed it once @maxatdetroit

src/stable/components/ProfileCard/ProfileCard.js
```diff
@@ -0,0 +1,82 @@
...
+</style>
+<a class="profile-card">
+  <img class="profile-image" alt="Profile Image">
+  <div class="profile-details">
+    <div class="name-container">
+      <slot name="name"></slot>
+      <span class="chevron">›</span>
+    </div>
+    <slot name="title-primary"></slot>
+    <slot name="title-secondary"></slot>
```
maxatdetroit: Let's make this a single slot called 'title'. If the person has multiple titles, the user can just insert multiple spans into the slot.  E.g.
```
<cod-profile-card src="https://placehold.co/400" href="https://example.com">
    <span slot="name">Jane Doe</span>
    <span slot="title">Frontend Engineer</span>
    <span slot="title">Frontend Developer</span>
  </cod-profile-card>
```

src/stable/components/ProfileCard/ProfileCard.js
```diff
@@ -0,0 +1,82 @@
...
+
+  connectedCallback() {
+    this._updateImage(this.src);
+    this._updateLink(this.href);
+  }
+
+  _updateImage(src) {
+    const img = this.shadowRoot.querySelector('.profile-image');
+    if (img) {
+      img.src = src || 'default-profile-image.jpg'; // Fallback image if src is invalid
```
maxatdetroit: Let's not use an arbitrary fallback filename. Just let it be an empty string in that case.

src/stable/components/ProfileCard/ProfileCard.js
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  _updateLink(newValue) {
+    const card = this.shadowRoot.querySelector('.profile-card');
+    if (card) {
+      if (newValue) {
+        card.href = newValue;
+        card.target = '_blank';
+        card.rel = 'noopener noreferrer';
```
maxatdetroit: These should be exposed as attributes on the component API (similar to href and src).

src/stable/components/ProfileCard/ProfileCard.js
```diff
@@ -0,0 +1,82 @@
...
+      } else {
+        card.removeAttribute('href');
```
maxatdetroit, line 100: You'll also have to remove the other attributes (target, rel), right?

src/stable/components/ProfileCard/ProfileCard.scss
```diff
@@ -0,0 +1,71 @@
...
+  font-weight: 400;
+  margin-bottom: 5px;
+}
+
+::slotted([slot='title-secondary']) {
+  font-size: 1rem;
+  line-height: 100%;
+  letter-spacing: 0%;
+  font-weight: 400;
+}
```
maxatdetroit: Recall what we discussed for slotted text styling: let's not introduce the text styles into the component style sheets, let the light DOM determine the text styles. But for illustration of how to style that text, move the text styles into a `<style>` object in the storybook story instead.

src/stable/docs/ProfileCard.mdx
```diff
@@ -0,0 +1,92 @@
...
+</doc-slots-table>
+
+## HTML Attributes / JS Properties
+
+Supported HTML attributes. If the attribute is reflected in a JS property, that attribute will have a 'reflects' tag next to the name.
+
+<doc-js-properties-table
+data={JSON.stringify([
+{"name": "href", "description": "A link to the page or site containing the person's information. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#href", "type": "string", "default": "-", "reflects": false},
+{"name": "src", "description": "The url to the proper page or site. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source", "type": "string", "default": "-", "reflects": false}
```
maxatdetroit: The description seems incorrect. Should be a URL to an image asset, not a page, right?

src/stable/docs/ProfileCard.mdx
```diff
@@ -0,0 +1,92 @@
...
+  data={JSON.stringify([])}>
+</doc-css-parts-table>
+
+## Dependencies
+
+<doc-dependencies-list data={JSON.stringify([])}></doc-dependencies-list>
+
+## Accessibility
+
+- **Images (`src`)**: When a profile image is included using the `src` attribute, it should be accompanied by an `alt` attribute describing the person (e.g., `alt="Photo of Jane Doe"`). This ensures screen readers can describe the image content.
```
maxatdetroit, line 90: We need to add the `alt` attribute to this component so it can be added correctly to the image element in the component as described here.

src/stable/components/ProfileCard/ProfileCard.js
```diff
@@ -0,0 +1,82 @@
...
+template.innerHTML = `
+<style>
+${styles}
+</style>
+<a class="profile-card">
+  <img class="profile-image" alt="Profile Image">
+  <div class="profile-details">
+    <div class="name-container">
+      <slot name="name"></slot>
+      <span class="chevron">›</span>
```
maxatdetroit: We should use the bootstrap icon chevron right instead here: https://icons.getbootstrap.com/icons/chevron-right/

maxatdetroit: It would also be nice if the chevron moved a bit to the right on hover to indicate it's clickable.

maxatdetroit (changes requested):
Just some minor tweaks at this point but still needs some changes before the component API is stable and ready to merge in.

src/stable/components/ProfileCard/ProfileCard.js
sreidthomas: Okay when I remove the -primary and -secondary its all the same thing so the secondary title shows but as whatever the -primary was. Shouldn't there be a distinction between the two @maxatdetroit:
image

maxatdetroit: Yes, there should be a distinction. You can leave separate controls for primary and secondary titles in the storybook example, but the underlying slots should both be `slot=title` and both the titles should be placed into that slot.

src/stable/docs/ProfileCard.mdx
sreidthomas: Do you mean change the description for both the href and src @maxatdetroit?

src/stable/components/ProfileCard/ProfileCard.js
```diff
@@ -0,0 +1,109 @@
...
+</style>
+<a class="profile-card">
+  <img class="profile-image" alt="Profile Image">
+  <div class="profile-details">
+    <div class="name-container">
+      <slot name="name"></slot>
+      <span class="chevron"></span>
+    </div>
+    <slot name="title"></slot>
+    <slot name="title"></slot>
```
maxatdetroit: You just need one slot for title. If someone tries to slot multiple elements, they get added to the same slot which is what we want. When I said we want to differentiate, I meant in storybook where we have two separate titles exposed as controls, but they both go into the same slot "under the hood".

src/stable/components/ProfileCard/ProfileCard.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  _updateLink() {
+    const card = this.shadowRoot.querySelector('.profile-card');
+    if (card) {
+      if (this._href) {
+        card.href = this._href;
+        card.target = this._target || '_blank';
+        card.rel = this._rel || 'noopener noreferrer';
```
maxatdetroit, line 98: Would you please remove all the fallback default values in the component? 

When we introduce default parameters, people using the component tend to rely on them and it makes it very hard to change the internal API in the future without breaking how people use the component. It's easier to maintain if we have no defaults, which forces people using the component to explicitly define the component behavior each time they use it.

maxatdetroit (changes requested):
This is the last set of feedback, couple minor things inline. Also:
- Could you make it so that the chevron animation happens on hover of any part of the link, not just the name text? 


https://github.com/user-attachments/assets/27ad9d43-60d5-4702-a902-916448701b02

PR #332, opened 2025-04-17
Update GovBanner Component

### LINKED ISSUE #328: Improvements to gov banner
A few improvement requests:
- [ ] The down chevron should rotate 180 degrees when the dropdown opens/closes to indicate the link now closes/opens the dropdown
- [ ] The display of the content container should use an ease-in-out animation instead of abruptly appearing
- [ ] The info section should be have a [`part` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/part) enabling it to be styled from the light DOM. Specifically, I should be able to override the left and right margins of the content so that the edges line up with the main content of the site where the banner is used.

Example of how I found these issues when using the component on a real site:

https://github.com/user-attachments/assets/f3f911cb-3590-4852-8fea-27fcf0830acd

sreidthomas:
### **GovBanner needed some updates to improve styles and behavior**

- [x] The down chevron should rotate 180 degrees when the dropdown opens/closes to indicate the link now closes/opens the dropdown
- [x] The display of the content container should use an ease-in-out animation instead of abruptly appearing
- [x] The info section should be have a [`part` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/part) enabling it to be styled from the light DOM. Specifically, I should be able to override the left and right margins of the content so that the edges line up with the main content of the site where the banner is used.


jedgar1mx (commented):
The child element should not have the `aria-expanded` attribute

src/stable/components/GovBanner/GovBanner.js
```diff
@@ -21,12 +21,12 @@ template.innerHTML = `
...
             </svg>
             </span>
-          </button>     
+          </button>    
       </div>
     </div>
   </header>
-  <div id="content" class="content-container" hidden>
-    <div class="info-section">
+  <div id="content" class="content-container">
```
maxatdetroit: This is the div that has the padding, so we need the CSS part to be defined on this div. Let's name the part `content` instead of `info-section`. 

Also, could we remove the next div (`.info-section`) entirely? This `#content` div should be enough to house the info items, right?

src/stable/components/GovBanner/GovBanner.scss
```diff
@@ -63,71 +54,69 @@
...
   width: 100%;
   border-top: 1px solid #dee2e6;
+  max-height: 0;
+  overflow: hidden;
+  transition:
+    max-height 0.3s ease-in-out,
+    padding 0.3s ease-in-out;
+  padding: 0 1rem;
+  &.visible {
+    max-height: 300px; /* Adjust this value based on your content height */
```
maxatdetroit: It'd be better to set `max-height` to `none` here which is the initial value (without any custom CSS). That way, if the content is larger than 300px tall, the box will expand to fit (at least while visible).

maxatdetroit (changes requested):
Just a couple small tweaks requested. Nice work.

maxatdetroit (changes requested):
I'm not sure what changed but the animation is now very jerky and the background color of the content area is incorrect on the left/right. Would you please correct those things?


https://github.com/user-attachments/assets/532c6778-08b2-49f6-bdb5-6d2e0126054f

sreidthomas:
https://github.com/user-attachments/assets/01753072-2122-4608-a270-4b245b1f4676

maxatdetroit (changes requested):
@sreidthomas could you merge in dev and resolve the conflicts? Sorry for the conflicts. I had to make some nontrivial changes in #350 to support a feature for the site launch this week.

PR #344, opened 2025-06-10
feature.LinkTray

### LINKED ISSUE #337: Create LinkTray component
## LinkTray

This task is to create a LinkTray component. Here's a single LinkTray (on mobile):

Image

Example of grid of LinkTray's (on desktop):

Image

### Detail Spec

- The LinkTray should have two slots: one slot for the tray title and another slot for the links.
- The LinkTray should have a CSS part so users can target its body to resize the tray.
- If the LinkTray is thin but sufficiently long (see mobile example above), the links should be arranged as a single column. If the LinkTray is resized so that it is wide but not very tall (see desktop example above), then the links should be forced into (at most) three columns of equal height.

SBBlee:
First draft of linktray, unfinished

image

### DIFF
--- a/src/stable/components/Linktray/Linktray.js
+++ b/src/stable/components/Linktray/Linktray.js
@@ -53,7 +53,7 @@ class Linktray extends HTMLElement {
   _fetchLinks() {
     const slot = this.shadowRoot.querySelector('slot[name="tray-link"]');
     if (!slot) return;
-    
+
     const wrapLinks = () => {
       const assignedElements = slot.assignedElements();
 
@@ -61,9 +61,9 @@ class Linktray extends HTMLElement {
         if (element.tagName === 'A') {
           const li = document.createElement('li');
           li.setAttribute('slot', 'tray-link');
-          element.removeAttribute('slot'); 
+          element.removeAttribute('slot');
           element.parentNode.insertBefore(li, element);
-          
+
           const chevron = document.createElement('span');
           chevron.classList.add('chevron');
           chevron.innerHTML = `
@@ -72,15 +72,15 @@ class Linktray extends HTMLElement {
             </svg> `;
           element.appendChild(chevron);
           li.appendChild(element);
-      } else if (element.tagName === 'LI') {
+        } else if (element.tagName === 'LI') {
           if (element.querySelector('a')) {
             element.setAttribute('slot', 'tray-link');
           } else {
             element.remove();
           }
-          } else {
+        } else {
           element.remove();
-          }
+        }
       });
     };
 
--- a/src/stable/components/Linktray/Linktray.js
+++ b/src/stable/components/Linktray/Linktray.js
@@ -5,11 +5,11 @@ template.innerHTML = `
   background: #f2f2f2; 
   border-left: 5px solid #feb70d; 
   padding: 20px 10px; 
+  width: 100%;
 
   ul {
     list-style-type: none;
     display: grid;
-    margin: 0 auto;
     gap: 1rem;
     margin: 1rem -2rem;
       @media (min-width: 600px) {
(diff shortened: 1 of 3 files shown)


src/stable/components/Linktray/Linktray.js
```diff
@@ -0,0 +1,80 @@
...
+          
+          const link = element.outerHTML;
+          
+          li.textContent = link;
+          
+          const chevron = document.createElement('span');
+          chevron.innerHTML = `
+          <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-chevron-right" viewBox="0 0 16 16">
+  <path fill-rule="evenodd" d="M4.646 1.646a.5.5 0 0 1 .708 0l6 6a.5.5 0 0 1 0 .708l-6 6a.5.5 0 0 1-.708-.708L10.293 8 4.646 2.354a.5.5 0 0 1 0-.708"/>
+</svg> `;
```
maxatdetroit: I noticed in the inspector that none of the slotted elements are wrapped in `li` tags, even though that's what is supposed to happen here. 

The reason is that we're creating the `li` element at line 62 here, but we're not inserting it into the component shadowDOM at any point.

image

To insert the `li` into the shadowDOM, we'd have to do something like this:

```js
element.parentNode.insertBefore(li, element);
```

That would basically tell the browser: find the parent (which happens to be the component root in this case `cod-linktray`) of the slotted element (the `a` element), and insert the `li` right before it. I pasted the rendered HTML for reference below.

```html
<cod-linktray class="linktray">

  <span slot="tray-title"> Mayor's Office </span>

  <a slot="tray-link" href="https://www.example.com">Mayor's Office
  </a><a slot="tray-link" href="https://www.example.com">DPD
  </a><a slot="tray-link" href="https://www.example.com">City Clerk
  </a><a slot="tray-link" href="https://www.example.com">HRD
  </a><a slot="tray-link" href="https://www.example.com">DWSD
  </a><a slot="tray-link" href="https://www.example.com">DPW

  </a><a slot="tray-link" href="https://www.example.com">Mayor's Office
  </a><a slot="tray-link" href="https://www.example.com">DPD
  </a><a slot="tray-link" href="https://www.example.com">City Clerk
  </a><a slot="tray-link" href="https://www.example.com">HRD
  </a><a slot="tray-link" href="https://www.example.com">DWSD
  </a><a slot="tray-link" href="https://www.example.com">DPW

</a></cod-linktray>
```

maxatdetroit: For now, I wouldn't worry about getting the content of the links to looks quite right, or the chevron to appear. Just try to get the story to render with the links wrapped in `li` elements.

src/stable/components/Linktray/Linktray.js
```diff
@@ -0,0 +1,91 @@
...
+        chevron.classList.add('chevron');
+        chevron.innerHTML = `
+          <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-chevron-right" viewBox="0 0 16 16">
+            <path fill-rule="evenodd" d="M4.646 1.646a.5.5 0 0 1 .708 0l6 6a.5.5 0 0 1 0 .708l-6 6a.5.5 0 0 1-.708-.708L10.293 8 4.646 2.354a.5.5 0 0 1 0-.708"/>
+          </svg> `;
+        li.appendChild(chevron);
+
+        // place li at the at the top of ul
+        const trayLinks = this.shadowRoot.querySelector('#tray-links');
+        trayLinks.insertBefore(li, trayLinks.children[0]);
```
maxatdetroit: This isn't quite right. This is taking the slotted `a` elements, moving them out of the light DOM slot, and moving them *into* the shadow DOM (and out of the slot).

Notice how there's nothing inside the `<slot name=tray-link>` slot:

image

We want the slotted elements to remain in the light DOM slot, but wrapped in `li` elements.

To keep the elements in the light DOM, we need to add the `slot=link-tray` attribute to the `li` tag created, and then append the `li` elements to the element `cod-linktray` instead of appending to an element in the shadowRoot like `'#tray-links`.

You can see an example of how this is done at lines 88-94 for the section navigation component: https://github.com/CityOfDetroit/COD-Design-System/blob/d908d2e63170f634630051934a4685efb37815bf/src/stable/components/SectionNavigation/SectionNavigation.js#L88-L94

Notice how the resulting HTML looks for the section navigation with the `li` elements inside the `<slot name=nav-items>`:

image

